PR #3, opened 2020-02-04
OP-118

npasquetto:
The issue [OP-118](https://openhospital.atlassian.net/browse/OP-118) is intended to expose a subset of OH modules as REST API.

This PR concerns only the Patient module, but it could be useful as:
* an example to implement other modules
* a basic module with which other modules interact
* a set of API that could be immediately used by [the frontend SPA](https://github.com/informatici/openhospital-ui)


mwithi:
Although it compiles and run with the latest changes in https://github.com/informatici/openhospital-core/pull/38 I'm not able to test successfully the API Authentication (/auth/login) nor the Swagger Authentication on an empy DB (admin:admin) and therefore also other API are untestable for me.

Any advice on Authentication process?

Note: OH-core artifact should remain 1.9

pom.xml
```diff
@@ -80,7 +80,7 @@
 		<dependency>
 			<groupId>org.isf</groupId>
 			<artifactId>OH-core</artifactId>
-			<version>1.9</version>
+			<version>2.0-SNAPSHOT</version>
```
mwithi: OH-core should remain 1.9 for compliance with other components

npasquetto:
Hi @mwithi!

> 
> Any advice on Authentication process?
> 

The User Agent must provide credentials to APIs with _basic access authentication_ method. 

So, the `/auth/login` route is only intended for login form, but if you put right credentials in the swagger-ui "**Try it now**" section of this route, you should get a 200 HTTP response code.

To test other routes, you need to click on "**Authorize**" button, fill the authentication form, confirm it (**don't expect a response at this stage!**) and then test any other method.

pom.xml
npasquetto: Ummm...Are you sure about this choice? 
How can we check which core version is delivered with a specific OS version?

mwithi:
@npasquetto just discovered the that authentication has no problem, only the way I externalized the connection settings (commit https://github.com/informatici/openhospital-api/commit/81ff29f1165d247f2474df1b98760af0e8d89194) brakes the connection itself, that's the reason why the authentication fails. 

So I'm currenlty working on it. Thanks

pom.xml
mwithi: Not sure I understood the question, the OH core will be the same for all OS versions and all components until we don't abandon the OH gui component, do you agree?

npasquetto: Sorry, typo in my comment, i meant: 

> How can we check which core version is delivered with a specific O**H** version

Each version of OH could use a different version of OH-core.
For exemple:

OH 1.9.4 -> OH-core 1.9
OH 1.9.10 -> OH-core 2.0
OH 1.9.11 -> OH-core 2.0
OH 1.9.15 -> OH-core 2.1
OH-api 2.0 -> OH-core 2.0

mwithi: Yes, but at the moment we don't have yet a system for that and I think we should puslish everything on Maven Central one day. Meanwhile, one must have the source code of each component in order to compile, and we never had OH-core 2.0 up to now, nor a set a rules for the individual versions' increment.

The only rule we had up to now is on the overall application version, that is:
Version 1.X.Y where Y increments on small changes not involving the DB, X increments for changes involving the DB.

But I'd like to discuss it further in the next call, that we are about to schedule.

PR #5, opened 2020-02-17
Op 127 improve exceptions handling (#2)

mwithi:
* OP-127 response contract api #time 2d
* OP-127 resource bundle to manage exception error messages #time 2h
* OP-127 OH custom exception handling management #time 2h
Co-authored-by: Alessandro Domanico <alessandro.domanico@yahoo.it>


mwithi:
Need to be done again

PR #12, opened 2020-03-17
OP-159 Bundle UI and API together for smooth build and packaging

paurav-munshi:
- Make UI as a sub modue of the API and add to src/main/openhospital
- Modify pom file to build UI along with API code.
- Bundle UI and API in same executable war file.


mwithi:
WIP

PR #13, opened 2020-03-20
OP-115 module from A to D

gildasdongmo:


src/main/java/org/isf/accounting/dto/BillDTO.java
```diff
@@ -0,0 +1,180 @@
...
+	
+	@NotNull
+	@ApiModelProperty(notes = "Bill balance", example="1500", position = 9)
+	private Double balance;
+	
+	@NotNull
+	@ApiModelProperty(notes = "user name who create the bill", example="admin", position = 10)
+	private String user;
+
+	private volatile int hashCode = 0;
```
npasquetto: Maybe at this stage we can avoid to use hashCode in DTOs.
We use them just as data transfert object.

src/main/java/org/isf/accounting/rest/BillController.java
```diff
@@ -0,0 +1,449 @@
...
+	@Autowired
+	protected PriceListManager priceListManager;
+	
+	@Autowired
+	protected PatientBrowserManager patientManager;
+
+	private final Logger logger = LoggerFactory.getLogger(BillController.class);
+
+	public BillController(BillBrowserManager billManager) {
+		this.billManager = billManager;
```
npasquetto: This constructor is not needed as the manager is autowired.

src/main/java/org/isf/accounting/rest/BillController.java
```diff
@@ -0,0 +1,449 @@
...
+    }
+	
+	/**
+	 * search all the {@link Bill}s associated to the passed {@link BillPayments}.
+	 * @param payments the {@link BillPayments} associated to the bill to retrieve.
+	 * @return a list of {@link Bill} associated to the passed {@link BillPayments} or <code>null</code> if an error occurred.
+	 * @throws OHServiceException
+	 */
+	@PostMapping(value = "/bills/search/by/payments", produces = MediaType.APPLICATION_JSON_VALUE)
+	public ResponseEntity<List<BillDTO>> searchBillsByPayments(@RequestBody ArrayList<BillPaymentsDTO> paymentsDTO) throws OHServiceException {
```
npasquetto: Could we use List instead of ArrayList as parameter type?

src/main/java/org/isf/disctype/rest/DischargeTypeController.java
```diff
@@ -0,0 +1,107 @@
...
+
+	@PutMapping(value = "/dischargetypes/{code}", produces = MediaType.APPLICATION_JSON_VALUE)
+	ResponseEntity<String> updateDischargeTypet(@PathVariable String code,
+			@RequestBody DischargeTypeDTO admissionTypeDTO) throws OHServiceException {
+		logger.info("Update dischargetypes code:" + code);
+		DischargeType admt = getObjectMapper().map(admissionTypeDTO, DischargeType.class);
+		admt.setCode(code);
+		if(!discTypeManager.codeControl(code)) 
+			throw new OHAPIException(
+					new OHExceptionMessage(null, "discharge type not found!", OHSeverityLevel.ERROR));
```
npasquetto: I think that the codeControl verify the presence of another entity with the same code. 
So it is not useful in an update context.

src/main/java/org/isf/disctype/rest/DischargeTypeController.java
```diff
@@ -0,0 +1,107 @@
...
+			admtCreated = admtFounds.get(0);
+		if (!isCreated || admtCreated == null) {
+			throw new OHAPIException(
+					new OHExceptionMessage(null, "discharge type is not created!", OHSeverityLevel.ERROR), HttpStatus.INTERNAL_SERVER_ERROR);
+		}
+		return ResponseEntity.status(HttpStatus.CREATED).body(admtCreated.getCode());
+	}
+
+	@PutMapping(value = "/dischargetypes/{code}", produces = MediaType.APPLICATION_JSON_VALUE)
+	ResponseEntity<String> updateDischargeTypet(@PathVariable String code,
```
npasquetto: Why do you ask for entity data and for entity code in two different parameteres?
It isn't easier to read the entity first (with getDischargeTypes() or a new search method) and then POST to this method only the updated DischargeTypeDTO?

npasquetto:
Hi guys, great job! I'm checking the code.

I've added some notes, and I have a question about the admission module DTOs: why did you have created 3 different DTOs, rather than just create one and evaluate the presence or not of the IDs in the nested DTOs?

src/main/java/org/isf/accounting/rest/BillController.java
tsognong: OK It's right. I am correcting it now.

src/main/java/org/isf/accounting/dto/BillDTO.java
tsognong: It's right. I am correcting it.

gildasdongmo:
> Hi guys, great job! I'm checking the code.
> 
> I've added some notes, and I have a question about the admission module DTOs: why did you have created 3 different DTOs, rather than just create one and evaluate the presence or not of the IDs in the nested DTOs?

gildasdongmo:
> Hi guys, great job! I'm checking the code.
> 
> I've added some notes, and I have a question about the admission module DTOs: why did you have created 3 different DTOs, rather than just create one and evaluate the presence or not of the IDs in the nested DTOs?

OK, I'll keep only one DTO but in this case I'll have to test all the mandatory fields manually since @valid won't be able to be used anymore, and also the model mapper doesn't map the patient field when all the object is not sent, to solve this I'll pass the patient id in request param.

src/main/java/org/isf/accounting/rest/BillController.java
tsognong: OK.

gildasdongmo:
Hi @npasquetto ,
we have made the necessary corrections.

src/main/java/org/isf/admission/rest/AdmissionController.java
```diff
@@ -0,0 +1,509 @@
...
+
+	/**
+	 * 
+	 * @param searchTerms
+	 * @param admissionRange
+	 * @param dischargeRange
+	 * @return
+	 * @throws OHServiceException
+	 */
+	@GetMapping(value = "/admissions/admitted/patient", produces = MediaType.APPLICATION_JSON_VALUE)
```
mwithi, line 136: I would suggest URL "/admissions/admittedPatients' for this one with pagination like in PatientController

src/main/java/org/isf/admtype/rest/AdmissionTypeController.java
```diff
@@ -0,0 +1,107 @@
...
+	protected AdmissionTypeBrowserManager admtManager;
+
+	private final Logger logger = LoggerFactory.getLogger(AdmissionTypeController.class);
+
+	public AdmissionTypeController(AdmissionTypeBrowserManager admtManager) {
+		this.admtManager = admtManager;
+	}
+
+	@PostMapping(value = "/admissiontypes", produces = MediaType.APPLICATION_JSON_VALUE)
+	ResponseEntity<String> newAdmissionType(@RequestBody AdmissionTypeDTO admissionTypeDTO) throws OHServiceException {
```
mwithi, line 45: no modifiers for some methods?

npasquetto:
> Only I'm thinking if it is not the case to use singular words for mapping URLs (`patient/` instead of `patients/` and `admission/` instead of `admissions/`)

Hi @mwithi in general it's better to use plural, as a convention. For reference: https://stackoverflow.com/questions/6845772/rest-uri-convention-singular-or-plural-name-of-resource-while-creating-it

src/main/java/org/isf/admission/rest/AdmissionController.java
gildasdongmo: Hi @mwithi I updated the URL, and the core didn't consider pagination for this method.

gildasdongmo: > Is there any reason why some methods from core have not been mapped?
> 
>     * getUsedWardBed(String wardId)
> 
>     * getNextYProg()
> 
> 
> probably there is plan to replace them by other business logic?
about this, I don't understand what this is all about!

mwithi: Ok, I've verified, and the pagination will be addressed by OP-161 👍

mwithi:
> > Only I'm thinking if it is not the case to use singular words for mapping URLs (`patient/` instead of `patients/` and `admission/` instead of `admissions/`)
> 
> Hi @mwithi in general it's better to use plural, as a convention. For reference: https://stackoverflow.com/questions/6845772/rest-uri-convention-singular-or-plural-name-of-resource-while-creating-it

Honestly I personally agree with the [comment](https://stackoverflow.com/a/33824071/5309541) and with the approved [answer](https://stackoverflow.com/a/338433/5309541) of a very similar discussion [page](https://stackoverflow.com/questions/338156/table-naming-dilemma-singular-vs-plural-names) (curiously they have comparable number of votes), also better explained by another [comment](https://stackoverflow.com/a/5841297/5309541) on the same page, that is the most ever voted (1786) among both pages, regardless the choosen answers.

The main reason I like singular compared to plural is that the former never change (like pre-existing DB naming standard), the second can be irregular.

Anyway, it's all about standards, not immutable rules, so what it is important is to choose one and stick to it. I also think that "standard" doesn't mean "better" or "logic", but that's it...

src/main/java/org/isf/admission/rest/AdmissionController.java
mwithi: > About this, I don't understand what this is all about!

`getUsedWardBed(String wardId)` is to know the beds occupation in a ward
`getNextYProgr()` is to query the database for next progressive number (within the Year) for an Admission

gildasdongmo: Okay, the methods have been mapped.

mwithi: Thanks! I will review it in no time!

mwithi: You used the same mapping, I suggest:
@GetMapping(value = "/admissions/getNextProgressiveIdInYear"
@GetMapping(value = "/admissions/getBedsOccupationInWard"

src/main/java/org/isf/admtype/rest/AdmissionTypeController.java
mwithi: is it ok to not have modifieds on some methods?

src/main/java/org/isf/admission/rest/AdmissionController.java
gildasdongmo: Okay, that's fixed.

PR #15, opened 2020-03-24
OP-120 modules from V to X

andreafalco1991:
The issue should contain this modules:
-vaccine
-vactype
-video -> (to develop)
-visits
-ward
-xmpp ->( i don't know if use object MessageListener is correct , probably not for ReactUI)


andreafalco1991:
vaccine: has problems on update and delete operations, maybe due to CORE
vactype: work correctly
video: is not to develop

mwithi:
> vaccine: has problems on update and delete operations, maybe due to CORE
> vactype: work correctly
> video: is not to develop

Hi @andreafalco1991 could you please provide more info about the problem?

andreafalco1991:
Hi @mwithi vaccineBrowserManager give me a DataIntegrityViolationException during delete or update of a vaccine?
I tried to create this vaccine:
```
{
  "code": "100",
  "description": "BCG",
  "vaccineType": {
    "code": "P",
    "description": "Pregnat"
  }
}
```
and update it with:
```
{
  "code": "100",
  "description": "BCGGG",
  "vaccineType": {
    "code": "P",
    "description": "Pregnat"
  }
}
```
but it give me that error.

Is it the expected behavior?

This is the error:
**org.springframework.dao.DataIntegrityViolationException: could not execute statement; SQL [n/a]; constraint [null]; nested exception is org.hibernate.exception.ConstraintViolationException: could not execute statement**

mwithi:
sure it is not then expected behavior, from 'gui' it works fine.
Anyway from a fast testing on your branch, the mvn clean install fails with:
`org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'xmppController' defined in file [C:\Users\Nanni\git\openhospital-api\target\classes\org\isf\xmpp\rest\XmppController.class]: Unsatisfied dependency expressed through constructor parameter 0: No qualifying bean of type [org.isf.xmpp.manager.Interaction] found for dependency [org.isf.xmpp.manager.Interaction]: expected at least 1 bean which qualifies as autowire candidate for this dependency. Dependency annotations: {}; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type [org.isf.xmpp.manager.Interaction] found for dependency [org.isf.xmpp.manager.Interaction]: expected at least 1 bean which qualifies as autowire candidate for this dependency. Dependency annotations: {}` 

did you change also something on the 'core' accordingly?

andreafalco1991:
It's my fault, i've just pushed a fix for it.

mwithi:
yeah now it build successfully!

I will review that along with the other PRs under OP-6 and its subtasks.

andreafalco1991:
Removed the xmpp module it should not be necessary as written in the comment of Emerson Castaneda inside issue https://openhospital.atlassian.net/browse/OP-6

mwithi:
Should we always think about a Response Body also for boolean/undefined original methods' return types?

andreafalco1991:
> Should new controllers be registered in /org/isf/config/SecurityConfig.java?

Yes, it should be registered under securityConfig but i don't know who can call the api, probably just admin



> Should we always think about a Response Body also for boolean/void original methods' return types?

I don't think that it's necessary return something if the client don't  use it.
It's better manage only the errors if they are presents otherwise if it is necessary i can return something, it's not a problem.

mwithi:
> Yes, it should be registered under securityConfig but i don't know who can call the api, probably just admin

At the moment only "admin" 

> I don't think that it's necessary return something if the client don't use it.
It's better manage only the errors if they are presents otherwise if it is necessary i can return something, it's not a problem.

I'd foster the discussion on [[Confluence] Rest API Controller Pattern](https://openhospital.atlassian.net/wiki/spaces/OHC/pages/769622123/Rest+API+Controller+pattern)

PR #17, opened 2020-04-04
OP-118 Modules from O to P

npasquetto:
We have introduced a modification to the initial pattern. 

- We have removed the `@DTO` annotation and the `ControllerAdvice` that, serializing the response object _after_ the controller execution, prevent SpringFox to generate a cohesive OpenAPI json specification.
+ We have introduced an AbstractController with the aim of avoid the rewriting of mapping methods in each controller, make the implementation of new controllers simplier.  

➡ More datails on [Jira](https://openhospital.atlassian.net/browse/OP-118)


bpraveen:
> I liked the idea about the abstract controller
> I have added ~#PR18~ [#PR1](https://github.com/intesys/openhospital-api/pull/1) directly on your fork with required adjustments for passing all tests and keeping all controllers compatibles with my original test suite proposal.
> Thanks

Hi Emecas, Since your latest PR#1 is dependent on this PR#17 OHApiAbstractController, please let me know if we can review this after PR#17 is merged.

npasquetto:
> Later on, I think the Abstract class will end up having more than desired functionality/GOD class and eventually having specific controller extends not-required methods.
> 
> If we can delay this decision it is better IMHO and for now use interface and later on refactor to Abstract class after most of our controllers are implemented. Please verify if this makes sense.
> 
> Making AbstractController for the sake of Mapper functionality does not make much sense at this point to me. But I'm new to this project and not aware of all of the discussion.

Does this make sense! I think about it. Thanks.

mwithi:
Discussion open at https://openhospital.atlassian.net/wiki/spaces/OHC/pages/769622123/Rest+API+Controller+pattern

npasquetto:
> Later on, I think the Abstract class will end up having more than desired functionality/GOD class and eventually having specific controller extends not-required methods.
> 
> If we can delay this decision it is better IMHO and for now use interface and later on refactor to Abstract class after most of our controllers are implemented. Please verify if this makes sense.
> 
> Making AbstractController for the sake of Mapper functionality does not make much sense at this point to me. But I'm new to this project and not aware of all of the discussion.

Ok @bpraveen , I've understand, I totally agree with you. My aim was to:
1 - avoid code repetition for marshal/unmarshal
2 - decouple from the serialization library implementation

We can reach target 2 with a simple  Interface/Implementation approach, but have you got any idea for point 1? Or we can ignore this at this stage?

npasquetto:
> Discussion open at https://openhospital.atlassian.net/wiki/spaces/OHC/pages/769622123/Rest+API+Controller+pattern

Maybe it's better if we discuss about it and report only the outcome, to avoid confusion on wiki pages, WDYT @mwithi ?

mwithi:
> Maybe it's better if we discuss about it and report only the outcome, to avoid confusion on wiki pages, WDYT @mwithi ?

I agree that the Wiki should host only the outcome, the one linked is a discussion page indeed. IMHO the discussion should be public if started by the community, the outcome clear and widely agreed, so no confusion can be generated, especially if all topics are treated publicly and no top-bottom decisions are taken.

bpraveen:
@npasquetto as mentioned in slack as well, as long as we don't expose the fields from Abstract Controiller as part of Json Response(DTO) we are still fine with it, we can make an internal refactor without worrying about backwards compatibility pertaining to this. 

So we can go ahead with merging this.

npasquetto:
Hi all and **thanks for feedbacks**! following suggestions I merged the PR by @emecas  that move the bind between Controller and Mapper from abstract class to implementation (and other minor fixes). This way also the @bpraveen hint to don't make the abstract controller a _GOD Object_ is met.

I've aligned the PR to the upstream, to made it mergable, we could also align the pattern on every controller, but IMO due to the huge amount of code involved we need to merge this PR ASAP to avoid conflict with other tasks and to enable some others to proceed. 

Note: the build fail due to an unaivalability of the bundlessgeo maven repo (https://repo.boundlessgeo.com/main/) that it's not involved in this task.

npasquetto:
> * Any particular reason  why one controller, out of the scope of this PR (modules from O to P) was included under `OHApiAbstractController` hierarchy, concretely: `DiseaseController`, and the others eleven controllers included here were not? what would be the approach about these controllers regarding to the pattern?

No reason at all, all rest controllers are going to be aligned to the pattern.

emecas:
@npasquetto reformulating my last question, should not we include the changes for the controllers in pending here as well? since the introduction of the abstract class is happening under this PR? or there is a reason to delay it?

Actually i started doing this tasks and created WIP-PR#2 in your fork-repo. Also i could complete the task if you want it. I would need some clarifications for the controllers that involve more than a DTO class like: BillController

Thanks

mwithi:
> 1. avoid code repetition for marshal/unmarshal

Question: if in then future we need to use the same mapper code outside of a controller, we could not use it, so the code would be definitely not reused or, even worse, repeated somewhere else?

> 2. decouple from the serialization library implementation

If we want to customize the mapping for only one model/controller we would end up changing it or providing it for all controllers

Also, as pointed by @emecas , some controllers deal with different models, so why to bind them to a single one? What to do with other models in the same controller? Still we have to map them in some way, probably repeating the code at point 1...

I think we'll define a different pattern with everyone's contribution at the discussione page, keeping in mind what we want to achieve; who wants to partecipate is very welcome. After the pattern will be approved (I hope shortly), it will be also documented and everybody will adopt it in their reciprocal work.

> Note: the build fail due to an unaivalability of the bundlessgeo maven repo (https://repo.boundlessgeo.com/main/) that it's not involved in this task.

The problem was raised under OP-140 and should be partially solved now (build success in openhospital-core)

mwithi:
Superseded by https://github.com/informatici/openhospital-api/pull/47

PR #21, opened 2020-04-17
OP-159 | Build API with UI to create a single bundle

paurav-munshi:


mwithi:
> Is it related to the script from the other PR? I'm a little bit lost here.

yes it is, the https://github.com/informatici/openhospital/pull/103 requires changes in "api" and "ui" components

mwithi:
Since the release process is totally changed (GitHub Action + makefile, Maven modules) the whole issue should be rewritten.

PR #22, opened 2020-04-22
Resolving SFL4J multiple bindings

mwithi:
Resources:
http://www.slf4j.org/codes.html#multiple_bindings
http://www.slf4j.org/codes.html#StaticLoggerBinder


mwithi:
Changes already on 'develop'

PR #24, opened 2020-04-30
Making PatientControllerTest compatible to new pattern

emecas:
- completing task listed on #23 
 - fixing configuration problem in #23

### DIFF
--- a/src/main/java/org/isf/patient/rest/PatientController.java
+++ b/src/main/java/org/isf/patient/rest/PatientController.java
@@ -45,8 +45,9 @@ public class PatientController {
 
 	private final Logger logger = LoggerFactory.getLogger(PatientController.class);
 
-	public PatientController(PatientBrowserManager patientManager) {
+	public PatientController(PatientBrowserManager patientManager,PatientMapper patientMapper) {
 		this.patientManager = patientManager;
+		this.patientMapper = patientMapper;
 	}
 
     /**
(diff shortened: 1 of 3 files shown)


emecas:
> 
> 
> Running mvn clean install that runs also tests I receive:
> `[02/mai/2020 17:07:05] [:] ERROR - Application startup failed org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'modelMapperConfig': Unsatisfied dependency expressed through field 'blobToByteArrayConverter': No qualifying bean of type [org.isf.shared.mapper.converter.BlobToByteArrayConverter] found for dependency [org.isf.shared.mapper.converter.BlobToByteArrayConverter]: expected at least 1 bean which qualifies as autowire candidate for this dependency. Dependency annotations: {@org.springframework.beans.factory.annotation.Autowired(required=true)}; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type [org.isf.shared.mapper.converter.BlobToByteArrayConverter] found for dependency [org.isf.shared.mapper.converter.BlobToByteArrayConverter]: expected at least 1 bean which qualifies as autowire candidate for this dependency. Dependency annotations: {@org.springframework.beans.factory.annotation.Autowired(required=true)}`

OK, but only by putting `@Autowired` back for the  `ModelMapper` in  `GenericMapper` is not going to fix that problem, it is a configuration issue, you need to add as well `@Component` annotation to  `BlobToByteArrayConverter` and `ByteArrayToBlobConverter`.

On the other hand, for keeping the unit testing  independent 23 add a conditional instantiation for the mapper in the constructor.

Added cd25cfb

mwithi:
> you need to add as well `@Component` annotation to `BlobToByteArrayConverter` and `ByteArrayToBlobConverter`

This is true, I'm going to add those (I don't get why Spring was ok without them).

PR #25, opened 2020-05-10
OP-6 Fix patient tests on new proposal pattern

vicusotto:
Fix patient tests on new proposal pattern

### DIFF
--- a/src/test/java/org/isf/patient/rest/PatientControllerTest.java
+++ b/src/test/java/org/isf/patient/rest/PatientControllerTest.java
@@ -31,16 +31,19 @@ import org.isf.patient.model.Patient;
 import org.isf.patient.test.TestPatient;
 import org.isf.shared.exceptions.OHAPIException;
 import org.isf.shared.exceptions.OHResponseEntityExceptionHandler;
+import org.isf.shared.mapper.converter.BlobToByteArrayConverter;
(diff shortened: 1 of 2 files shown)


src/test/java/org/isf/patient/rest/PatientControllerTest.java
```diff
@@ -61,18 +64,21 @@
 	@Mock
     private PatientBrowserManager patientBrowserManagerMock;
 	
-	@Autowired
-	private static PatientMapper mapper;
-	
+	private static PatientMapper mapper = new PatientMapper();
```
emecas, line 67: Naming this field `patientMapper` as well as removing the `static` would be clearer.

`private PatientMapper patientMapper = new PatientMapper();`

Then you could fix  `PatientDTOHelper` `setup` method like this:

```
		public static PatientDTO setup(PatientMapper patientMapper) throws OHException{
			Patient patient = PatientHelper.setupPatient();
			return patientMapper.map2DTO(patient);
		}
```
and provide `patientMapper` for all the calls of `PatientDTOHelper` `setup` method. 

Note: that in a different pull request I've already moved the helper classes to independent files, having an static field to be able to use the object in a nested class in a single place is not enough reason to add coupling.

vicusotto: Yes I agree.
Thanks!

PR #45, opened 2020-07-17
Op-119 module from S to U

MartinezX21:


src/main/java/org/isf/sms/rest/SmsController.java
```diff
@@ -0,0 +1,103 @@
...
+		return ResponseEntity.ok(true);
+	}
+	
+	/**
+	 * Deletes the specified {@link Sms}
+	 * @param smsDTOList
+	 * @return <code>true</code> if the sms is deleted
+	 * @throws OHServiceException
+	 */
+	@PostMapping(value = "/sms/delete", produces = MediaType.APPLICATION_JSON_VALUE)
```
emecas, line 94: any particular reason for avoiding  @DeleteMapping here?

mwithi: @MartinezX21 any update? could you please also resolve the conflict in src/main/java/org/isf/supplier/dto/SupplierDTO.java
?

uni2grow: @emecas The fact is that the core provides no way to find SMS by ID or any unique identifier, thus the entire SMS object should be sent to be deleted. I read that a lot of clients are unable to send a DELETE request with a body like Angular (HttpClient). That's why I used @PostMapping for this special case.

MartinezX21: @emecas The fact is that the core provides no way to find SMS by ID or any unique identifier, thus the entire SMS object should be sent to be deleted. I read that a lot of clients are unable to send a DELETE request with a body like Angular (HttpClient). That's why I used @PostMapping for this special case.

emecas: > 
> 
> @emecas The fact is that the core provides no way to find SMS by ID or any unique identifier, thus the entire SMS object should be sent to be deleted. I read that a lot of clients are unable to send a DELETE request with a body like Angular (HttpClient). That's why I used @PostMapping for this special case.

I think this is fair enough for the SMS/ID case, I made a search and there is a lot of discussion regarding to this subject with 2 valid positions 1) specification is unclear so why not?,  but 2) adding a body on DELETE is semantically meaningless.

Thanks @MartinezX21

PR #56, opened 2020-08-20
Replace Travis CI with Github Actions

pviotti:
Due to various reason discussed previously (stability, uniformity, etc) we adopt GitHub Actions.


.github/workflows/maven.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Checkout core
-      run: git clone --depth=50 --branch=develop https://github.com/informatici/openhospital-core.git openhospital-core
+      run: git clone --depth=50 --branch=OP-6-implement-the-new-rest-api https://github.com/informatici/openhospital-core.git openhospital-core
```
mwithi, line 16: this is really needed only on the related [branch](https://github.com/informatici/openhospital-api/tree/OP-6-implement-the-new-rest-api) on this repo, so I think we can keep 'develop' here.

pviotti: yeah that's what I was trying to figure out. Develop branch doesn't seem to be working with any core branch AFAIK. We could checkout the core branch named as the api branch we're in... but that might just be a temporary workaround for that new-api branch.

PR #57, opened 2020-08-21
[OP-241] Update Delete VaccineController

emecas:
Removing `RequestBody` parameter from `@DeleteMapping deleteVaccine` 

Fix OP-241


mwithi (approved):
LGTM

PR #59, opened 2020-08-29
OP-116 modules from e to l

vicusotto:
First pr for modules e to l new rest api, includes: exams, examrows, examtypes, examinations, hospitals, laboratories


src/main/java/org/isf/hospital/rest/HospitalController.java
```diff
@@ -0,0 +1,75 @@
...
+    protected HospitalBrowsingManager hospitalBrowsingManager;
+    @Autowired
+    private HospitalMapper hospitalMapper;
+
+    public HospitalController(HospitalBrowsingManager hospitalBrowsingManager, HospitalMapper hospitalMapper) {
+        this.hospitalBrowsingManager = hospitalBrowsingManager;
+        this.hospitalMapper = hospitalMapper;
+    }
+
+    @PutMapping(value = "/hospitals/{code:.+}", produces = MediaType.APPLICATION_JSON_VALUE)
```
mwithi, line 37: At the moment the application doesn't manage more than one hospital, should we consider only UPDATE api?

src/main/java/org/isf/config/SecurityConfig.java
```diff
@@ -134,6 +134,36 @@ protected void configure(final HttpSecurity http) throws Exception {
 				.antMatchers(HttpMethod.DELETE, "/wards/**").hasAuthority("admin")
 				.antMatchers(HttpMethod.PATCH, "/wards/**").hasAuthority("admin")
 				.antMatchers(HttpMethod.GET, "/wards/**").hasAnyAuthority("admin", "guest")
+				.antMatchers(HttpMethod.POST, "/exams/**").hasAuthority("admin")
```
emecas, line 137: We need it for all controllers.

We could use a regular expression to reduce the number of `antMatchers` required.

See: https://docs.spring.io/spring-security/site/docs/4.1.1.RELEASE/apidocs/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.html#regexMatchers-org.springframework.http.HttpMethod-java.lang.String...-

src/main/java/org/isf/exam/rest/ExamController.java
```diff
@@ -0,0 +1,120 @@
...
+        boolean isCreated = examManager.newExam(exam);
+        if (!isCreated) {
+            throw new OHAPIException(new OHExceptionMessage(null, "Exam is not created!", OHSeverityLevel.ERROR));
+        }
+        return ResponseEntity.ok(exam.getCode());
+    }
+
+    @PutMapping(value = "/exams/{code:.+}", produces = MediaType.APPLICATION_JSON_VALUE)
+    public ResponseEntity updateExams(@PathVariable String code, @RequestBody ExamDTO updateExam) throws OHServiceException {
+
```
MartinezX21, line 64: @vicusotto Is it necessary to take the parameter exam _code_ here? Since lines 65-67 are used to check whether the code passed corresponds to the code in the _updateExam_ object and throw an exception if not. I think this data is redundant.

src/main/java/org/isf/examination/dto/PatientExaminationDTO.java
```diff
@@ -0,0 +1,185 @@
...
+    private Integer pex_hgt;
+
+    private Integer pex_diuresis;
+
+    private String pex_diuresis_desc;
+
+    private String pex_bowel_desc;
+
+    private String pex_note;
+
```
MartinezX21, line 48: Is there a particular reason for not annotating the other properties?

src/main/java/org/isf/examination/rest/ExaminationController.java
```diff
@@ -0,0 +1,163 @@
...
+
+        examinationBrowserManager.saveOrUpdate(patientExamination);
+
+        return ResponseEntity.status(HttpStatus.CREATED).body(patientExamination.getPex_ID());
+    }
+
+    @PutMapping(value = "/examinations/{id}", produces = MediaType.APPLICATION_JSON_VALUE)
+    @ResponseStatus(HttpStatus.OK)
+    public ResponseEntity updateExamination(@PathVariable Integer id, @RequestBody PatientExaminationDTO dto) throws OHServiceException {
+        if (dto.getPex_ID() != id) {
```
MartinezX21, line 64: In think the _id_ param here is redundant.

src/main/java/org/isf/exatype/rest/ExamTypeController.java
```diff
@@ -0,0 +1,95 @@
...
+        if (!created) {
+            throw new OHAPIException(new OHExceptionMessage(null, "ExamType type not created!", OHSeverityLevel.ERROR));
+        }
+
+        return ResponseEntity.status(HttpStatus.CREATED).body(true);
+    }
+
+    @PutMapping(value = "/examtypes/{code:.+}", produces = MediaType.APPLICATION_JSON_VALUE)
+    public ResponseEntity updateExamType(@PathVariable String code, @RequestBody ExamTypeDTO updateExamType) throws OHServiceException {
+
```
MartinezX21, line 55: I think the _code_ param here in a redundant information.

src/main/java/org/isf/hospital/rest/HospitalController.java
```diff
@@ -0,0 +1,75 @@
...
+    public ResponseEntity updateHospital(@PathVariable String code, @RequestBody HospitalDTO hospitalDTO) throws OHServiceException {
```
MartinezX21, line 38: The _code_ param here is not necessary.

src/main/java/org/isf/lab/rest/LaboratoryController.java
```diff
@@ -0,0 +1,242 @@
...
+    public LaboratoryController(LabManager laboratoryManager, PatientBrowserManager patientBrowserManager, ExamBrowsingManager examManager, LaboratoryMapper laboratoryMapper, LaboratoryRowMapper laboratoryRowMapper) {
+        this.laboratoryManager = laboratoryManager;
+        this.patientBrowserManager = patientBrowserManager;
+        this.examManager = examManager;
+        this.laboratoryMapper = laboratoryMapper;
+        this.laboratoryRowMapper = laboratoryRowMapper;
+    }
+
+    @PostMapping(value = "/laboratories", produces = MediaType.APPLICATION_JSON_VALUE)
+    public ResponseEntity<Integer> newLaboratory(@RequestBody LaboratoryDTO laboratoryDTO, @RequestBody(required = false) List<String> labRow) throws OHServiceException {
```
MartinezX21, line 64: There is quite a bit of discussion on the Internet about using multiple @RequestBody in parameter has the same method. It is not recommended to do so.
What you need to do:
- Create an object that includes the laboratoryDTO and labRow objects,
e.g.
`class LabAndRowsDTO{
    LaboratoryDTO laboratoryDTO;
    List<String> labRow;
}`
- Use this object in the _newLaboratory_ method and access its properties to extract the required information.
`public ResponseEntity<Integer> newLaboratory(@RequestBody LabAndRowsDTO laboratoryAndRowsDTO) OHServiceException throws {...}`

src/main/java/org/isf/lab/rest/LaboratoryController.java
```diff
@@ -0,0 +1,242 @@
...
+
+        if (!inserted) {
+            throw new OHAPIException(new OHExceptionMessage(null, "Laboratory is not created!", OHSeverityLevel.ERROR));
+        }
+
+        return ResponseEntity.status(HttpStatus.CREATED).body(labToInsert.getCode());
+    }
+
+    @PostMapping(value = "/laboratories/insertList", produces = MediaType.APPLICATION_JSON_VALUE)
+    public ResponseEntity newLaboratory2(@RequestBody List<LaboratoryDTO> laboratoryDTOs, @RequestBody ArrayList<ArrayList<LaboratoryRowDTO>> labsRows) throws OHServiceException {
```
MartinezX21, line 99: Same remark as https://github.com/informatici/openhospital-api/pull/59/files#r482110726

src/main/java/org/isf/lab/rest/LaboratoryController.java
```diff
@@ -0,0 +1,242 @@
...
+        boolean inserted = laboratoryManager.newLaboratory2(labsToInsert, labsRowsToInsert);
+
+        if (!inserted) {
+            throw new OHAPIException(new OHExceptionMessage(null, "Laboratory is not created!", OHSeverityLevel.ERROR));
+        }
+        return ResponseEntity.status(HttpStatus.CREATED).body(true);
+    }
+
+    @PutMapping(value = "/laboratories/{code}", produces = MediaType.APPLICATION_JSON_VALUE)
+    public ResponseEntity updateLaboratory(@PathVariable Integer code, @RequestBody LaboratoryDTO laboratoryDTO, @RequestBody(required = false) List<String> labRow) throws OHServiceException {
```
MartinezX21, line 142: Same remark as https://github.com/informatici/openhospital-api/pull/59/files#r482110726

MartinezX21 (changes requested):

src/main/java/org/isf/hospital/rest/HospitalController.java
vicusotto: Do you mean to delete /hospitals (GET) and /hospitals/currencyCode (GET).
How could we retrieve hospitals from api?

src/main/java/org/isf/config/SecurityConfig.java
vicusotto: Yes, this method is very verbose for all controllers. 
IMHO regex probably introduce less code maintainability, what do you think?

src/main/java/org/isf/exam/rest/ExamController.java
vicusotto: I use to build PUT method with code as uri parameter. It seems a convention of Api Crud method. Found this [https://www.w3.org/Protocols/rfc2616/rfc2616.txt] HTTP standard docs, for the PUT method (9.6 PUT):

> The fundamental difference between the POST and PUT requests is
reflected in the different meaning of the Request-URI. The URI in a
POST request identifies the resource that will handle the enclosed
entity. That resource might be a data-accepting process, a gateway to
some other protocol, or a separate entity that accepts annotations.
In contrast, the URI in a PUT request identifies the entity enclosed
with the request -- the user agent knows what URI is intended and the
server MUST NOT attempt to apply the request to some other resource.
If the server desires that the request be applied to a different URI

Do you know it's a deprecated method usage?

mwithi (commented):
> Do you mean to delete /hospitals (GET) and /hospitals/currencyCode (GET).

No, I was referring only to the PUT{code:.+}, at the moment we can manage only one record for that table

src/main/java/org/isf/examination/dto/PatientExaminationDTO.java
vicusotto: You are right, I didn't know the meaning of all properties :-)

src/main/java/org/isf/lab/rest/LaboratoryController.java
vicusotto: I agree with you and updated pr with your advice.

src/main/java/org/isf/examination/rest/ExaminationController.java
```diff
@@ -0,0 +1,163 @@
...
+        this.patientBrowserManager = patientBrowserManager;
+    }
+
+    @PostMapping(value = "/examinations", produces = MediaType.APPLICATION_JSON_VALUE)
+    public ResponseEntity<Integer> newPatientExamination(@RequestBody PatientExaminationDTO newPatientExamination) throws OHServiceException {
+        String patName = newPatientExamination.getPatient().getFirstName() + " " + newPatientExamination.getPatient().getSecondName();
+        Patient patient = patientBrowserManager.getPatient(patName);
+        if (patient == null) {
+            throw new OHAPIException(new OHExceptionMessage(null, "Patient not exists!", OHSeverityLevel.ERROR));
+        }
```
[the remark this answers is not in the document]
vicusotto: Yes I think it's better but with DTO Model mapping, I created the exact mapping for PatientExamination model.
I updated pr with your advice. Do you prefer this implementation also for other controllers?

I didn't a specialized DTO only for put/post method but also for get, so get api will not retrieve examination with the entire patient information.
Do you think they are necessary and need a specialized dto for the get methods?

vicusotto:
> > Do you mean to delete /hospitals (GET) and /hospitals/currencyCode (GET).
> 
> No, I was referring only to the PUT{code:.+}, at the moment we can manage only one record for that table

I know it's a standard to provide PUT method with code/id in the URI, as I reported before from HTTP standard docs in @MartinezX21 comment [https://github.com/informatici/openhospital-api/pull/59#discussion_r484099127](url) 
Do you prefer regardless to delete code from path uri?

src/main/java/org/isf/hospital/rest/HospitalController.java
mwithi: Sorry, my mistake, I wrongly understood that notation was for an array of `code`

src/main/java/org/isf/exam/rest/ExamController.java
MartinezX21: I think you're right. It's true that this way of doing things involves additional work to double-check the concordance of identifiers, but it complies with standards.

MartinezX21 (approved):

MartinezX21:
It' s ok for me. 
Thanks @vicusotto

vicusotto:
> It' s ok for me.
> Thanks @vicusotto

Thanks for your review @MartinezX21

src/main/java/org/isf/examination/rest/ExaminationController.java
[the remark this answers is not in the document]
mwithi: Sorry, my bad, I misunderstood that notation `{code:.+}`

src/main/java/org/isf/config/SecurityConfig.java
mwithi: I think we can tackle this problem in a second step, @emecas and @vicusotto do you agree?

src/main/java/org/isf/examination/dto/PatientExaminationDTO.java
mwithi: Added with https://github.com/informatici/openhospital-api/pull/59/commits/56ca102a38431292987512aa5767ce773a5df91e

src/main/java/org/isf/config/SecurityConfig.java
emecas: yes , we can leave this for a different stage during API implementation, if new code that you are adding here to SecurityConfig.java  is not affecting or representing changes for the current and future test suite(s) implementation ( I don't think so), I would give an approve to this PR.

mwithi: > is not affecting or representing changes

I think so, anyway I'm going to test it as well

src/main/java/org/isf/examination/rest/ExaminationController.java
```diff
@@ -0,0 +1,156 @@
+package org.isf.examination.rest;
```
emecas, line 1: is not a @DeleteMapping required fr this controller/module?

src/main/java/org/isf/config/SecurityConfig.java
emecas: you are rigth, i just confirmed on my side too!!!

emecas (commented):
Subject about security configuration is clarified now.

Just an additional verification for ExaminationController, if a DeleteMapping is required or not

src/main/java/org/isf/examination/rest/ExaminationController.java
mwithi: it has been introduced in "core" after OP-6 branch, I don't know if it's better to add now or later after merge into develop, wdyt?

emecas: I would open a linked ticket/task with 2 subtasks to keep that work in pending on track and then merge this PR in order someone could start working on the respective test suite in parallel (OP-197) 

Task: Add delete mapping to ExaminationController

Subtask 1: sync op-6 branch with change on develop

subtask 2: once subtask 1 is completed, make the addition to the controller

mwithi: OK, so I resolve this conversation concerning this PR.

About the tasks, there's no need to "sync op-6 branch" (subtask1) if we postpone the work after OP-6 is completed, because its completion means it will be already merged into develop

emecas (approved):
Added future task: "OP-266 Add DeleteMapping in ExaminationController"

PR #62, opened 2020-09-08
OP-260 | Replace @RequestBody with  @PathVariable in @DeleteMapping WardController

emecas:
Signed-off-by: Emerson Castaneda <emecas@users.noreply.github.com>

### DIFF
--- a/src/main/java/org/isf/ward/rest/WardController.java
+++ b/src/main/java/org/isf/ward/rest/WardController.java
@@ -7,6 +7,7 @@ import org.isf.shared.exceptions.OHAPIException;
 import org.isf.utils.exception.OHServiceException;
 import org.isf.utils.exception.model.OHExceptionMessage;
 import org.isf.utils.exception.model.OHSeverityLevel;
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/ward/rest/WardController.java
```diff
@@ -7,6 +7,7 @@
 import org.isf.utils.exception.OHServiceException;
 import org.isf.utils.exception.model.OHExceptionMessage;
 import org.isf.utils.exception.model.OHSeverityLevel;
+import org.isf.vaccine.model.Vaccine;
```
mwithi, line 10: Is this really needed?

mwithi (approved):
Apart the unused import, it looks fine.

```
Results :

Tests run: 87, Failures: 0, Errors: 0, Skipped: 0
```

src/main/java/org/isf/ward/rest/WardController.java
emecas: > Is this really needed?

It is fixed now. Thanks

PR #70, opened 2020-09-19
FIX OP-277 | Ambiguous Handler Methods Mapped

emecas:


mwithi (commented):
It works fine, nevertheless I've realized that the method `org.isf.ward.manager.WardBrowserManager.maternityControl(boolean)` cannot work if bundles are not in the classpath (they are located in "gui") so it saves in the DB the following:

  | M | angal.admission.maternity | 234/52544 | 54324/5424 | maternity@stluke.org | 20 | 3 | 2 | 0 | 0 | 1 | 0 |   | 2020-09-19 21:13:37 |   | 2020-09-19 21:13:37 | 1
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --

Indeed, there are some bundles that are needed in "core" (like angal.sql.*) and others that are only needed in "gui", should we split it and bundle it in OH-core-<version>.jar or other suggestions?

Meanwhile we could just "hide" this method waiting for a solutions.

emecas:
Does the save record have any error(s) ?

I am trying to understand what you meant.

I am testing against `CORE ` `OP-6-implement-the-new-rest-api` branch (d90de44d4).

I did not find any dependence from `GUI` at `org.isf.ward.managerWardBrowserManager`.

`org.isf.generaldata.MessageBundle` is defined on `CORE`, but it has this an import of `javax.swing.JOptionPane`.

For `maternityControl(boolean)` at `WardBrowserManager` is not existent test at CORE, neither for called methods:  `getDefaultMaternityWard()`,  `newWard(Ward ward)` , or `validateWard(Ward ward, boolean insert)`

but call for isMaternityPresent() defined at  `org.isf.ward.service.WardIoOperations` is under testing at `org.isf.ward.test.Tests`  `testIoIsMaternityPresent` test.

mwithi:
1. yes, "angal.admission.maternity" is a key in the bundle files used in "gui", it depends by the language and should be translated to "MATERNITY" if the user language is English
2. yes, everything is ok (thus I don't know why is setting a property for the JOptionPane never used
3. the method is used in GUI, so it must be present in CORE, so to hide in API because at the moment it needs bundles (that are in GUI)

...or any other solution about bundles. Actually we are not caring about languages in API, are we?

emecas:
> 1. yes, "angal.admission.maternity" is a key in the bundle files used in "gui", should be translated to "MATERNITY" if the user's language is English, it depends by the GeneralData.LANGUAGE setting.

OK... I can see now what you are referring, i was looking around code dependencies but it is about configuration key-value items

> 2. yes, everything is OK (thus I don't know why is setting a property for the JOptionPane never used

Then it can be removed

> 3. the method is used in GUI, so it must be present in CORE, so to hide in API because at the moment it needs bundles (that are in GUI)
> 
> ...or any other solution about bundles. 

As a quick solution, instead of hide functionality, we could break that coupling between API and GUI by creating a new key-value item or moving/copying it to an accessible place.

As a general solution, what about a configuration that we could inject for all modules which required it,  even adopting a central place for all system and using profiles?

>Actually we are not caring about languages in API, are we?
> 

We have not faced that need because API testing implemented until now is only unit testing controllers and mappings by mocking persistency (managers and operation repositories implicitly)

mwithi:
> As a general solution, what about a configuration that we could inject for all modules which required it, even adopting a central place for all system and using profiles?

Yeah, it's what I was suggesting, in the first comment: splitting the bundle in order to have bundles for CORE and others for GUI.
TBH the "technology" of properties bundle is meant and used only for the GUI and CORE should never use it, but it is not so... the `MessageBundle.getMessage(String key)` method is used 531 times in `CORE\develop`, especially for _validations error messages_ in **managers**, but also for _SQL errors handling_ (`angal.sql.*`) in **services** and for `localized datasets` between managers and **gui** (e.g. `MaritalStaus` and `Profession` in Patient).

So for CORE, we could think to have CORE-related ones in another bundle folder in that repo, that will also serve when OH-core is loaded as dependency. 
For API, we faced the problem once, and there was a suggested [proposal](https://github.com/vicusotto/openhospital-api/commit/883d421e1622f20b06459d5a6ba4ea64ef97021d#diff-600376dffeb79835ede4a0b285078036) (by @vicusotto) to create a [bundle repository](https://github.com/vicusotto/openhospital-bundle) too, that anyway would host ALL translations, also the GUI-related ones that would be out of scope, but still a suitable solution (also for the Transifex platform we are using)
For UI, there is a discussion [here](https://openhospital.atlassian.net/wiki/spaces/OHC/pages/619544602/UI+-+Translations) if to use the same technology (properties files) or to adopt a more suitable one for front-end developments; in the first case make sense, again, to have a common 'bundle' repository for all components, in the second one not really...

For settings we have kind of same problem of bundles: we decided to keep them in the GUI, but now that also other components need them, we should do something about it. If we decide to move them into CORE, if I'm not wrong, for bundles we need to use different names for the properties files, otherwise they would be replaced by the GUI in the classpath; in contrary for the settings a replacement in the classpath could be fine but I'm not sure of that, we did it for `database.properties` [here](https://github.com/informatici/openhospital-core/blob/655accaf6c2ab2715d31f3eabb19db8dd80a4adc/src/main/resources/applicationContext.xml#L34).

mwithi:
I've created a new issue OP-291 for the discussion above, I'm going to merge this as is because the problem will be solved elsewhere anyway

PR #76, opened 2020-10-29
OP-335 Lombok for logging changes

dbmalkovsky:
Added `@Slf4j` annotation in every file that had a Slf4j import statement even if there were no actual logging calls in the file.  This also required changing the static field from `logger` to Lombok's preference of `log`.


mwithi:
LGTM.

mwithi (commented):
LGTM

PR #78, opened 2020-10-30
Chore: use a parameterized log message

dbmalkovsky:
Change non-constant string concatenations used as arguments to SLF4J methods. Non-constant concatenations will be evaluated at runtime even when the logging message is not logged; this can negatively impact performance.  A parameterized log message will not be evaluated when logging is disabled.

Also replaced a `System.out.println` call with a logging call instead.


emecas (approved):
LGTM

Thanks

mwithi (approved):
Thanks a lot!

PR #81, opened 2020-10-30
Chore: With JDK7 type arguments can be replaced with diamond type <>

dbmalkovsky:


mwithi:
Benefit of this PR is just cosmetic, right?

dbmalkovsky:
Yes you could consider it cosmetic.   Starting with JDK7 you don't have to specify the same information it two places.   So it also cuts down on typing.

dbmalkovsky:
Just resolved the two file conflicts.

PR #82, opened 2020-10-31
Op 335 lombok getter setter

dbmalkovsky:
Had to change at least one variable name (`isPatient` to `patientBill` in `BillDTO.jav)a` so the generation of the boolean method didn't conflict.

Made all `getHasCode()` methods consistent by adding `@ApiModelProperty(hidden= true)`.

Made all `getLock()` methods consistent by adding `@ApiModelProperty(hidden= true)`.

Noticed most `lock` variables were `int`; but some where `Integer`; didn't change them.

Notice that some files had `private volatile int hashCode = 0;` while others had `private int hashCode = 0;`; left them as they were defined.

Changed some method names in the tests to match the generated code:

- getBillDTO() -> getBill()
- getBillItemsDTO() -> getBillItems()
- getBillPaymentsDTO() -> getBillPayments()

Removed unneeded `super()` calls.

Added missing `@Override` annotations as needed.

### DIFF
--- a/src/main/java/org/isf/accounting/dto/BillDTO.java
+++ b/src/main/java/org/isf/accounting/dto/BillDTO.java
@@ -29,11 +29,7 @@ import org.isf.patient.dto.PatientDTO;
 
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
-import lombok.Getter;
-import lombok.Setter;
 
-@Getter
-@Setter
 @ApiModel(description = "Class representing a bill")
 public class BillDTO {
 	
@@ -61,7 +57,7 @@ public class BillDTO {
 	
 	@NotNull
 	@ApiModelProperty(notes = "Is bill belongs to a patient?", example="true", position = 5)
-	private boolean isPatientBill;
+	private boolean patientTrue;
 	
 	@NotNull
 	@ApiModelProperty(notes = "patient name", example="Mario Rossi", position = 6)
@@ -82,5 +78,108 @@ public class BillDTO {
 	@NotNull
 	@ApiModelProperty(notes = "user name who create the bill", example="admin", position = 10)
 	private String user;
-	
+
+	public Integer getId() {
+		return this.id;
+	}
+
+	public PatientDTO getPatient() {
+		return this.patient;
+	}
+
+	public Integer getListId() {
+		return this.listId;
+	}
+
+	public Date getDate() {
+		return this.date;
+	}
+
+	public Date getUpdate() {
+		return this.update;
+	}
+
+	public boolean isList() {
+		return this.isList;
+	}
+
+	public String getListName() {
+		return this.listName;
+	}
+
+	public boolean isPatientTrue() {
+		return this.patientTrue;
+	}
+
+	public String getPatName() {
+		return this.patName;
+	}
+
+	public String getStatus() {
+		return this.status;
+	}
+
+	public Double getAmount() {
+		return this.amount;
+	}
+
+	public Double getBalance() {
+		return this.balance;
+	}
+
+	public String getUser() {
+		return this.user;
+	}
+
+	public void setId(Integer id) {
+		this.id = id;
+	}
+
+	public void setPatient(PatientDTO patient) {
+		this.patient = patient;
+	}
+
+	public void setListId(Integer listId) {
+		this.listId = listId;
+	}
+
+	public void setDate(Date date) {
+		this.date = date;
+	}
+
+	public void setUpdate(Date update) {
+		this.update = update;
+	}
+
+	public void setList(boolean isList) {
+		this.isList = isList;
+	}
+
+	public void setListName(String listName) {
+		this.listName = listName;
+	}
+
+	public void setPatientTrue(boolean isPatientBill) {
+		this.patientTrue = isPatientBill;
+	}
+
+	public void setPatName(String patName) {
+		this.patName = patName;
+	}
+
+	public void setStatus(String status) {
+		this.status = status;
+	}
+
+	public void setAmount(Double amount) {
+		this.amount = amount;
+	}
+
+	public void setBalance(Double balance) {
+		this.balance = balance;
+	}
+
+	public void setUser(String user) {
+		this.user = user;
+	}
 }
--- a/src/main/java/org/isf/accounting/dto/BillDTO.java
+++ b/src/main/java/org/isf/accounting/dto/BillDTO.java
@@ -61,7 +61,7 @@ public class BillDTO {
 	
 	@NotNull
 	@ApiModelProperty(notes = "Is bill belongs to a patient?", example="true", position = 5)
-	private boolean patientBill;
+	private boolean isPatientBill;
 	
 	@NotNull
 	@ApiModelProperty(notes = "patient name", example="Mario Rossi", position = 6)
--- a/src/main/java/org/isf/accounting/dto/BillDTO.java
+++ b/src/main/java/org/isf/accounting/dto/BillDTO.java
@@ -29,7 +29,11 @@ import org.isf.patient.dto.PatientDTO;
 
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
+import lombok.Getter;
+import lombok.Setter;
 
+@Getter
+@Setter
 @ApiModel(description = "Class representing a bill")
 public class BillDTO {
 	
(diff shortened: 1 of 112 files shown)


src/main/java/org/isf/accounting/dto/BillDTO.java
```diff
@@ -57,7 +61,7 @@
 	
 	@NotNull
 	@ApiModelProperty(notes = "Is bill belongs to a patient?", example="true", position = 5)
-	private boolean isPatient;
+	private boolean patientBill;
```
mwithi, line 64: Are we sure of this? The core model has both:
```
	@NotNull
	@Column(name="BLL_IS_PAT")
	private boolean isPatient;
	
	@ManyToOne
	@JoinColumn(name="BLL_ID_PAT")
	private Patient billPatient;
```

src/main/java/org/isf/accounting/rest/BillController.java
```diff
@@ -131,9 +131,9 @@ public BillController(BillBrowserManager billManager, PriceListManager priceList
         	throw new OHAPIException(new OHExceptionMessage(null, "Price list not found!", OHSeverityLevel.ERROR));
         }
         
-        ArrayList<BillItems> billItems = new ArrayList<BillItems>(billItemsMapper.map2ModelList(newBillDto.getBillItemsDTO()));
+        ArrayList<BillItems> billItems = new ArrayList<BillItems>(billItemsMapper.map2ModelList(newBillDto.getBillItems()));
```
mwithi, line 134: Not clear to me: if we want to rename the `billItems` property into `billItemsDTO` will Lombok update also the related getters and setters automatically?

mwithi (commented):
Some questions in the comments.
Also, why [here](https://github.com/informatici/openhospital-api/pull/82/files#diff-9aa0cf994fe779d8ee4c139b100e1d38e4b4dadc803b93407e5c370a235cee60R40) we don't have annotations? Probably patient was the "missing" patientBill that has been put in [R64](https://github.com/informatici/openhospital-api/pull/82/files#diff-9aa0cf994fe779d8ee4c139b100e1d38e4b4dadc803b93407e5c370a235cee60R40)

src/main/java/org/isf/accounting/rest/BillController.java
dbmalkovsky: Yes that is true but most of the methods/properties did not use the "DTO" part so I was making them consistent.

src/main/java/org/isf/accounting/dto/BillDTO.java
dbmalkovsky: There was a name conflict in the generated method name and this was a way to resolve it.  I'll need to take a closer look.

src/main/java/org/isf/accounting/rest/BillController.java
mwithi: yes the intention was clear, I was just asking how Lombok works and helps (also) in refactoring because I don't know it 

(TBH I tried to refactor `billItems`->`billItemsDTO` in IDE Eclipse and getters in [R134](https://github.com/informatici/openhospital-api/pull/82/files#diff-0c44ffb6e41726bcf884d8c344783d40884e5470a453adf29e74bb8bd63dbabaR185) is not updated accordingly)

src/main/java/org/isf/accounting/dto/BillDTO.java
dbmalkovsky: @mwithi Upon closer examination if `isPatient` is left as it was these are the methods generated:
```
	public @NotNull boolean isPatient() {
		return this.isPatient;
	}

	public void setPatient(@NotNull boolean isPatient) {
		this.isPatient = isPatient;
	}
```

But the `PatientDTO patient` variable generates
````
	public void setPatient(PatientDTO patient) {
		this.patient = patient;
	}
````
If I use maven then the compile is clean but if I try in the IDE compiling the `BillController` file with the annotations it complains with:
```
java: incompatible types: boolean cannot be converted to org.isf.patient.dto.PatientDTO
```
It appears that it cannot resolve the fact that there are 2 method signatures for `setPatient`.  This is probably a bug in the IDE but leaving it that way makes development at best inconvenient. 

If I de-lombok the file (that is expand all the generated methods) inline and try compiling in the IDE then it works fine which seems to validate the issue with annotation processing in IntelliJ when only using the generated code from the annotations.

So I've pushed another change to make `patientBill` into the variable `isPatientBill`.  I guess will need to live with it.

File: src/main/java/org/isf/accounting/rest/BillController.java (hunk 2 above)
Comment (dbmalkovsky) [reply src/main/java/org/isf/accounting/rest/BillController.java:134]: The methods that are called are not changed automatically.  When the Lombok annotations are added behind the scenes the `getXXX`, `setXXX`, and maybe `isXXX` methods are generated.   Any place that calls the methods are not automatically updated.   That is still a manual process.

Comment (mwithi) [reply src/main/java/org/isf/accounting/rest/BillController.java:134]: I see... I think Lombok should cover this because otherwise the good of having removed boilerplate code (that before you was writing once and for all) is diminished by the absence of an important feature in IDEs that is the one of refactoring properties and methods automatically (and not manually). And it seems I'm foreshadowing what other people already experienced, like also the "[find references](https://stackoverflow.com/q/31675937/5309541)" feature (even more useful indeed).

Maybe we can suggest/document also the workarounds proposed by those same people?

Comment (dbmalkovsky) [reply src/main/java/org/isf/accounting/rest/BillController.java:134]: You are enumerating the reasons that I don't like Lombok.  All it does really is not have getters/setters in your code.   Given that most modern IDE's will automatically generate the code (and often with Javadoc) it really isn't saving any typing.   It is just too much (partial) magic for my taste.   If you haven't figured it out by now I tend to have opinions on many topics :)

Comment (dbmalkovsky) [reply src/main/java/org/isf/accounting/rest/BillController.java:134]: I'll be more than happy to undo the changes and create a new PR.

Comment (mwithi) [reply src/main/java/org/isf/accounting/rest/BillController.java:134]: Actually I was not deciding [alone that](https://github.com/informatici/openhospital-api/pull/16), but I totally agree with you now that I know what Lombok is and does.
Let me ask @pviotti if agrees to undo changes and also remove the plugin?

File: src/main/java/org/isf/accounting/dto/BillDTO.java (hunk 1 above)
Comment (mwithi) [reply src/main/java/org/isf/accounting/dto/BillDTO.java]: To avoid such situations I use to name the variables of **ACertainType** with the type's name but lowercase, so something like:
```
private PatientDTO patientDTO;
```

For booleans, taking into account what the IDEs normally do, it's to say adding to a given <property> the words "is" and "set" in this way:
```
getter -> is<Property>
setter -> set<Property>
```

I think **isPatient** should be renamed to something more self-explaining, like:
```
private boolean patientTrue;
...
getter -> isPatientTrue();
setter -> setPatientTrue(boolean patientTrue);
```

Otherwise the proposed changes already work fine for me 👍

src/main/java/org/isf/accounting/rest/BillController.java
dbmalkovsky: @pviotti Are you OK with not using Lombok?

pviotti: fine by me. I'd really love to remove some boilerplate code but if it comes at the price of hindering features of popular IDEs, then maybe not.

dbmalkovsky: OK; then the decision appears to be to remove Lombok; I'll create a new PR tomorrow that restores the logging and the `boilerplate` code.  Simple and straightforward is always the best approach IMHO.

emecas: Remove Lombok!!!  interesting and kind of radical solution :) 

@dbmalkovsky  @pviotti  @mwithi Would it be a temporal solution?

I am windering how the progresses about Lombok adoption is in the other modules CORE and GUI ?

dbmalkovsky: Just pushed the changes to remove Lombok in this project.

@mwithi I also mad the change to the method name as you suggested; now using `patientTrue`.

dbmalkovsky: @emecas This appears to be the only project that started to use Lombok.   The cost benefit ratio for using Lombok is just too high (IMHO) if you want to do easy refactoring and have complete control over project.

mwithi: > @mwithi I also made the change to the method name as you suggested; now using `patientTrue`.

Thanks a lot!

> if you want to do easy refactoring and have complete control over project.

yeah, and it seems a forgotten issues on their [roadmap](https://groups.google.com/g/project-lombok/search?q=refactoring)

mwithi:
Resolved conflict directly online, could you please verify the indentation and formatting? Thanks

dbmalkovsky:
Looks OK.

mwithi:
> Looks OK.

new conflicts here because of #81 😅

dbmalkovsky:
Resolved conflicts.

PR #83, opened 2020-11-01
Chore - remove unnecessary toString()

dbmalkovsky:
The method itself takes care of the conversion to a String object.

BTW, in the last case:
```
    @Override
    public String toString() {
        return "VisitDTO{" +
                ", patient=" + patient +
                ", date=" + date +
                ", note='" + note + '\'' +
                ", sms=" + sms +
```
the `patient` `toString()` method is not overridden so it will just object an object id.

### DIFF
--- a/src/main/java/org/isf/vaccine/rest/VaccineController.java
+++ b/src/main/java/org/isf/vaccine/rest/VaccineController.java
@@ -111,7 +111,7 @@ public class VaccineController {
      */
     @PostMapping(value = "/vaccines", produces = MediaType.APPLICATION_JSON_VALUE)
     public ResponseEntity newVaccine(@RequestBody VaccineDTO newVaccine) throws OHServiceException {
-	    log.info("Create vaccine: {}", newVaccine.toString());
+	    log.info("Create vaccine: {}", newVaccine);
         boolean isCreated;
         try {
              isCreated = vaccineManager.newVaccine(mapper.map2Model(newVaccine));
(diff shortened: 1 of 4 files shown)


src/main/java/org/isf/vaccine/rest/VaccineController.java
```diff
@@ -111,7 +111,7 @@ public VaccineController(VaccineBrowserManager vaccineManager, VaccineMapper vac
      */
     @PostMapping(value = "/vaccines", produces = MediaType.APPLICATION_JSON_VALUE)
     public ResponseEntity newVaccine(@RequestBody VaccineDTO newVaccine) throws OHServiceException {
-	    log.info("Create vaccine: {}", newVaccine.toString());
```
mwithi, line 114: here the `toString()` is actually overridden in `VaccineDTO` class, and I think with the only intention to have this log, but it doesn't seem to be a costant behaviour in OP-6 (there are many contributors) so what to do?

I would get rid of them (also existing the overridden methods) and maybe define a common pattern for all DTO classes (if really needed at a certain point).

Otherwise this PR should not remove what is actually overridden (and so not unnecessary)

mwithi (commented):
Question

src/main/java/org/isf/vaccine/rest/VaccineController.java
dbmalkovsky: You don't need to specify `toString()` explicitly to get the overridden behavior.  Given this code:
````
package org.isf;

public class MyString {

	public static void main(String[] args) {
		MyClass myClass = new MyClass();
		System.out.println("myClass: " + myClass);
		System.out.println("myClass.toString(): " + myClass.toString());
	}

	static class MyClass {

		int x = 1;

		@Override
		public String toString() {
			return "This is from toString()";
		}
	}
}
````

The output is:
````
myClass: This is from toString()
myClass.toString(): This is from toString()
````

mwithi: You are right, I got the point now.

mwithi:
> the `patient` `toString()` method is not overridden so it will just object an object id.

yeah, we'll live with it until a different need will be raised 😄

Thanks a lot for your work!

PR #88, opened 2020-11-19
Chore Only need general Exception defined

dbmalkovsky:
With method signature: `throws JsonProcessingException, Exception` there is no need for the `JsonProcessingException` given the more general `Exception`.

### DIFF
--- a/src/test/java/org/isf/agetype/rest/AgeTypeControllerTest.java
+++ b/src/test/java/org/isf/agetype/rest/AgeTypeControllerTest.java
@@ -80,7 +80,7 @@ public class AgeTypeControllerTest {
 	}
 
 	@Test
-	public void testGetAllAgeTypes_200() throws JsonProcessingException, Exception {
+	public void testGetAllAgeTypes_200() throws Exception {
 		String request = "/agetypes";
 
 		ArrayList<AgeType> results = AgeTypeHelper.genArrayList(5);
@@ -147,7 +147,7 @@ public class AgeTypeControllerTest {
 	}
 
 	@Test
-	public void testGetAgeTypeByIndex_200() throws JsonProcessingException, Exception {
+	public void testGetAgeTypeByIndex_200() throws Exception {
 		String request = "/agetypes/{index}";
 		int index = 10;
 		AgeType ageType = AgeTypeHelper.setup(index);
--- a/src/test/java/org/isf/disease/rest/DiseaseControllerTest.java
+++ b/src/test/java/org/isf/disease/rest/DiseaseControllerTest.java
@@ -99,7 +99,7 @@ public class DiseaseControllerTest {
 	}
 
 	@Test
-	public void testGetDiseasesOpdByCode_200() throws JsonProcessingException, Exception {
+	public void testGetDiseasesOpdByCode_200() throws Exception {
 		String request = "/diseases/opd/{typecode}";
 
 		String typeCode = "1";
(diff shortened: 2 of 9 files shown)


dbmalkovsky:
This was found using IntelliJ's Analyze tool.  The inspection that found this was `Multiple exceptions declared on test method`.  The subtext is
```
Reports JUnit test methods with more than one exception declared in the throws clause. Such a throws clause can be more concisely declared as: 
throws Exception
````

emecas (commented):
@dbmalkovsky , the particular change makes sense, but I am wondering if instead to generalize everything to use Exception, we could go deeper to add more tests and define/use our own or particular exception(s) in order to be able to difference through multiple potential errors, lets say categories related to: data parsing, absent/invalid data, persistence, etc. 

Also, it could be too much work. What do you think?

CC @mwithi

dbmalkovsky:
@emecas I'm not sure we need to do that; if we get an exception traceback it will indicate what the root problem is.

mwithi:
To throw detailed Exception(s) might be useful for calling classes, but... is it really important for @Test methods?

emecas:
I'm sorry, i am mixing related subjects here, please let me clarify, what I was referring before was more about test coverage. It totally depends of  what we want to cover with the tests, most of the tests I have added are only focused on testing the happy paths for each method in the controller.

For example for the first changed file here: `AgeTypeController`, the test coverage is like this:

image

Why?

Because tests are missing cover all the  5 possible error/exception conditions:

1.  [Error] when in `getAllAgeTypes()` ageTypeManager.getAgeType() returns and empty list
1.  [Error] when in  `updateAgeType(@Valid @RequestBody AgeTypeDTO ageTypeDTO)` the `code `in the body `ageTypeDTO` is empty
1.  [Exception] also when in `updateAgeType(@Valid @RequestBody AgeTypeDTO ageTypeDTO)` the call `ageTypeManager.updateAgeType(ageTypes)` returns `false`
1.  [Exception] when in `getAgeTypeCodeByAge(@RequestParam("age") int age)` the call `ageTypeManager.getTypeByAge(age)` returns `null`
1.  [Error] when in `getAgeTypeByIndex(@PathVariable int index)` the call `ageTypeManager.getTypeByCode(index)` returns `null`

image

For this particular PR, that is only focused on fixing Exception subject, but all methods in the  controller I'm using as example `AgeTypeController` throw `OHServiceException`, but actually `Exception` is only related to the method:  `perform(RequestBuilder requestBuilder)` from `MockMvc` class,  and methods `andDo(ResultHandler handler)` ,  `andExpect(ResultMatcher matcher) ` from `ResultActions` interface .

Note: here I just wanted clarify about coverage subject, but while I was working on an proposal for fixing this, I found other problem in Core, so i will follow up later, .

Thanks

dbmalkovsky:
Thanks for the clarification; it makes a lot of sense.   My view is that a test methods is basically an independent *main* program that  no other method is calling and is dependent on the Exception contact defined.   Adding `throws Exception` is just making the Java compiler and the semantics happy.  IMHO I think we are over thinking things.

emecas:
> 
> 
> Thanks for the clarification; it makes a lot of sense. My view is that a test methods is basically an independent _main_ program that no other method is calling and is dependent on the Exception contact defined. Adding `throws Exception` is just making the Java compiler and the semantics happy. IMHO I think we are over thinking things.

Yes, you are totally right, I just wanted to let you know what I think is relevant point about exceptions and testing coverage, but we can take care of that in a different task ... thanks

mwithi:
So I think we are happy with merging this 🙂 
Glad we had chance to clarify many things.
Thanks a lot to both!

PR #92, opened 2020-11-20
Chore Fix reference comparison

dbmalkovsky:
Found using SpotBugs (the successor to FindBugs).

The description of the issue:
````
Suspicious reference comparison
This method compares two reference values using the == or != operator, where the correct way to compare instances of this 
type is generally with the equals() method. It is possible to create distinct instances that are equal but do not 
compare as == since they are different objects. Examples of classes which should generally not be compared by
reference are java.lang.Integer, java.lang.Float, etc.
````

So I changed the `==` or `!=` into the corresponding version using `equals()`.

FYI @emecas


emecas (approved):
Lgtm, thanks

PR #95, opened 2020-11-23
Chore Remove dead store

dbmalkovsky:
Found with SpotBugs

Remove creation of `admissionType` as it is not used in the test method.

### DIFF
--- a/src/main/java/org/isf/ward/rest/WardController.java
+++ b/src/main/java/org/isf/ward/rest/WardController.java
@@ -131,7 +131,7 @@ public class WardController {
     public ResponseEntity newWard(@RequestBody WardDTO newWard) throws OHServiceException {
 	    LOGGER.info("Create Ward: {}", newWard);
         Ward wardCreated = wardManager.newWard(mapper.map2Model(newWard));
-        if (wardCreated != null) {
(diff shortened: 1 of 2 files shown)


dbmalkovsky:
@emecas Can you please confirm as it appears that you are the original author?  Thanks

dbmalkovsky:
Locally no tests failed but in the CI build 2 did fail:
````
Failed tests: 
  WardControllerTest.testNewWard_200:172 Range for response status value 400 expected:<SUCCESSFUL> but was:<CLIENT_ERROR>
  WardControllerTest.testUpdateWard_200:196 Range for response status value 400 expected:<SUCCESSFUL> but was:<CLIENT_ERROR>
````

Neither of those tests were the single test that I touched in this PR.   I think there have been some recent Ward changes in Core.

dbmalkovsky:
@mwithi I believe your fix for the ward tests was incorrect.  Please review my changes to that produced a clean local run and a clean CI build.  Probably this should be committed so others don't see the same failures.

mwithi (approved):
@dbmalkovsky yes, I noticed that just after but had no time to fix! Thanks!

PR #96, opened 2020-11-24
Chore Replace "size() == 0" with "isEmpty()"

dbmalkovsky:
Found with SonarLint which notes

````
Using Collection.size() to test for emptiness works, but using Collection.isEmpty() makes the code more
readable and can be more performant. The time complexity of any isEmpty() method implementation 
should be O(1) whereas some implementations of size() can be O(n).
````

Project had 57 occurrences of `isEmpty()` and 64 occurrences of `size() == 0` before the changes made in this PR.

Also added `logs` directory to .gitignore to simplify future comparisons.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,5 +24,9 @@ HELP.md
 /nbdist/
 /.nb-gradle/
 /build/
+
+
+
 *.log
 /oh-rest-api.log.*.gz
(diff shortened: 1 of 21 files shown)


mwithi (approved):
LGMT. Thanks!

PR #97, opened 2020-11-29
OP-376 Swagger 2 generated spec compliance fixes 

npasquetto:
https://openhospital.atlassian.net/browse/OP-376

Due to some mistakes mainly on SpringFox annotations and REST controllers paths definitions, the autogenerated api-docs has some errors against the Swagger 2 specification.To check that, I copy the generated api-docs (from /oh-api/v2/api-docs) and paste it in swagger editor. Those are the reported errors:

oh-opeapi-spec-errors

### DIFF
--- a/src/main/java/org/isf/stats/rest/ReportsController.java
+++ b/src/main/java/org/isf/stats/rest/ReportsController.java
@@ -21,8 +21,13 @@
  */
 package org.isf.stats.rest;
 
-import io.swagger.annotations.Api;
-import io.swagger.annotations.Authorization;
+import java.io.IOException;
+import java.net.MalformedURLException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+import javax.servlet.http.HttpServletRequest;
+
 import org.apache.poi.util.IOUtils;
 import org.isf.shared.exceptions.OHAPIException;
 import org.isf.stat.dto.JasperReportResultDto;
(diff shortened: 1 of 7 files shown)


src/main/java/org/isf/stats/rest/ReportsController.java
```diff
@@ -21,13 +21,9 @@
  */
 package org.isf.stats.rest;
```
dbmalkovsky, line 23: Please use the  IDE formatting files for in the `.ide-settings` directory.  The changes to the imports here do not match project standards:

1. static imports
2. java
3. javax
4. org
5. com
6. everything else

Thanks

npasquetto: Done!

mwithi (approved):
The new generated json (after converting to YAML)
image

Tests also ok:
```
Results :

Tests run: 105, Failures: 0, Errors: 0, Skipped: 0
```

npasquetto:
I've searched if there is some way to integrate this check in maven or in the pipeline but it seems that there isn't a _clean_ way to do this.

mwithi:
> I've searched if there is some way to integrate this check in maven or in the pipeline but it seems that there isn't a _clean_ way to do this.

That would be a very good shot indeed! The process should be:

1. get json from URL /v2/api-docs
2. call Swagger editor API for the check (eventually first convert to YAML)
3. parse the response and interrupt the pipeline

correct?

npasquetto:
Yes, also with a junit test for example, but "the swagger editor api" doesn't exists 😔

There is a way to replicate the validaton done by swagger editor in a Java method, but it is too tricky IMO.

I'm going to open and issue in backlog about that.

mwithi:
> I'm going to open and issue in backlog about that.

👍

PR #98, opened 2020-11-29
OP-378 Moved springfox auth settings from controllers to centralized config

npasquetto:
https://openhospital.atlassian.net/browse/OP-378


mwithi (changes requested):
Should be done the same on all controllers? 
image

npasquetto:
> Should be done the same on all controllers?

Absolutly, maybe I've done a wrong search to identify where the Authorization annotation was used 🤦‍♂️
Now should be ok.

I've also removed an unsed import from all classes.

mwithi (approved):

PR #100, opened 2020-12-29
OP-377 Implements JWT authentication

npasquetto:
The purpose of this PR is to replace the *Basic* Authentication mechanism with the *JWT* Authentication mechanism as a stateless, more secure, cookie independant solution.

**Spring Security**

Fully integrated in Spring Security, main components are:
* a **token provider** that generate the token
* a **login success handler** thet return the generated token on succesfully login
* a **security filter** that check and validate the token

**SpringFox**

* updated spring fox version (that need an update on spring boot and the addition of gson dependency)
* updated *SpringFoxConfig* class to support the JWT security schema and context

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -106,11 +106,6 @@
             <version>2.9.2</version>
         </dependency>
         <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
-            <version>20.0</version>
-        </dependency>
-        <dependency>
             <groupId>io.springfox</groupId>
             <artifactId>springfox-swagger-ui</artifactId>
             <version>2.6.1</version>
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>org.springframework.boot</groupId>
         <artifactId>spring-boot-starter-parent</artifactId>
-        <version>1.4.0.RELEASE</version>
+        <version>1.4.1.RELEASE</version>
         <relativePath/> <!-- lookup parent from repository -->
     </parent>
     <groupId>org.isf</groupId>
@@ -103,7 +103,12 @@
         <dependency>
             <groupId>io.springfox</groupId>
             <artifactId>springfox-swagger2</artifactId>
-            <version>2.6.1</version>
+            <version>2.9.2</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>20.0</version>
         </dependency>
         <dependency>
             <groupId>io.springfox</groupId>
@@ -114,6 +119,23 @@
             <groupId>org.glassfish</groupId>
             <artifactId>javax.el</artifactId>
             <version>3.0.1-b12</version>
+        </dependency>
+        <dependency>
+            <groupId>io.jsonwebtoken</groupId>
+            <artifactId>jjwt-api</artifactId>
+            <version>0.11.1</version>
+        </dependency>
+        <dependency>
+            <groupId>io.jsonwebtoken</groupId>
+            <artifactId>jjwt-impl</artifactId>
+            <version>0.11.1</version>
+            <scope>runtime</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.jsonwebtoken</groupId>
+            <artifactId>jjwt-jackson</artifactId>
+            <version>0.11.1</version>
+            <scope>runtime</scope>
         </dependency>
     </dependencies>
 
--- a/src/main/java/org/isf/security/LoginResponse.java
+++ b/src/main/java/org/isf/security/LoginResponse.java
@@ -1,3 +1,24 @@
+/*
+ * Open Hospital (www.open-hospital.org)
+ * Copyright © 2006-2020 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
+ *
+ * Open Hospital is a free and open source software for healthcare data management.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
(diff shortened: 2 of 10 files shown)


src/main/java/org/isf/security/LoginResponse.java
```diff
@@ -0,0 +1,23 @@
+package org.isf.security;
```
dbmalkovsky, line 1: Missing license header.

src/main/java/org/isf/security/OHSimpleUrlAuthenticationSuccessHandler.java
```diff
@@ -27,6 +27,8 @@
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
```
dbmalkovsky, line 30: per project import ordering standards this import should appear after the `org` ones

src/main/java/org/isf/security/jwt/JWTConfigurer.java
```diff
@@ -0,0 +1,21 @@
+package org.isf.security.jwt;
```
dbmalkovsky, line 1: Missing license header

src/main/java/org/isf/security/jwt/JWTFilter.java
```diff
@@ -0,0 +1,48 @@
+package org.isf.security.jwt;
```
dbmalkovsky, line 1: missing license header

src/main/java/org/isf/security/jwt/JWTFilter.java
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.util.StringUtils;
+import org.springframework.web.filter.GenericFilterBean;
+
+import javax.servlet.FilterChain;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import java.io.IOException;
```
dbmalkovsky, line 13: imports in wrong order; see in this repo `.ide-settings\eclipse\OpenHosptital,importorder`.   The order is:

1. static
2. java
3. javax
4. org
5. com
6. everything else

src/main/java/org/isf/security/jwt/TokenProvider.java
```diff
@@ -0,0 +1,98 @@
+package org.isf.security.jwt;
```
dbmalkovsky, line 1: Same two comments as above.

npasquetto:
Done some fixes about license and import order.

mwithi:
Very nice! Nevertheless, while I'm able to get and use the generated token from Postman with same CURL command, I'm not able to get it working in Swagger: login/ returns response code 0, and although I set it as authorization key, every api return same response. 

Same for you?

pom.xml
```diff
@@ -103,7 +103,12 @@
         <dependency>
             <groupId>io.springfox</groupId>
             <artifactId>springfox-swagger2</artifactId>
-            <version>2.6.1</version>
+            <version>2.9.2</version>
+        </dependency>
+        <dependency>
```
dbmalkovsky, line 108: I don't see were this dependency is imported; is it really used?

mwithi: @npasquetto do we?

dbmalkovsky: Just to be clear I was asking about the google.guava dependency at lines 109-110

mwithi: good point, I understood springfox-swagger2 indeed 😆

npasquetto:
> Very nice! Nevertheless, while I'm able to get and use the generated token with Postman using same CURL command, I'm not able to get it work in Swagger: `login/` returns `response code: 0`, and although I set it as authorization key, every `api/` using the token in the CURL command, return same response.
> 
> Same for you?

Yes, it was due to an error in `api.host` default value in SpringFox configuration, now it should to be ok.
Thanks @mwithi!

pom.xml
npasquetto: No: I've introduced google.guava dependency as a workaround for a bug in a previous version of springfox2 (https://github.com/springfox/springfox/issues/2232) that I've tested before the MR submission, but in the springfox2 version used in the MR, the bug it's been fixed, so the workaround (and the dependency) it's no longer needed.

mwithi:
> Yes, it was due to an error in `api.host` default value in SpringFox configuration, now it should to be ok.
> Thanks @mwithi!

Thanks a lot @npasquetto !!
Now the login is working fine in Swagger, but the token seems not been recognized using Swagger (with Postman the same token do work). 

Maybe I'm doing wrong?

image

image

npasquetto:
> > Yes, it was due to an error in `api.host` default value in SpringFox configuration, now it should to be ok.
> > Thanks @mwithi!
> 
> Thanks a lot @npasquetto !!
> Now the login is working fine in Swagger, but the token seems not been recognized using Swagger (with Postman the same token do work). 
> 
> Maybe I'm doing wrong?
> 
> image
> 
> image
> 
> 
> 
> 

In this version of swagger-ui you need to prepend the token with the 

'Bearer ' 

keyword (space included); this is needed by authorization filters to distinguish between credentials type - basic, bearer, others. 

Check out how Postman set the `Authorization` header on outgoing request.

BTW, I'm going to submit a PR to update the swagger-ui version, as in newer versions this behavior is managed.

mwithi:
> keyword (space included); this is needed by authorization filters to distinguish between credentials type - basic, bearer, others.

I was just wondering how it could guess it! Thanks, it works perfectly now!

> BTW, I'm going to submit a PR to update the swagger-ui version, as in newer versions this behavior is managed.

If I'm not wrong we were waiting for bumping this version (there's a dependabot PR about) because some legacy reasons due to Spring 1.x

PR #110, opened 2021-03-11
Chore: only need a primitive type

dbmalkovsky:
There is no need to use a wrapper version of boolean in this method.


mwithi (approved):
Thanks!

PR #111, opened 2021-03-14
Chore: unnecessary unboxing

dbmalkovsky:
The method `admissionManager.newAdmissionReturnKey(newAdmission)` returns an `int` and not an `Integer` thus making the test for null and `aId.intValue()` unnecessary.

`Unboxing`,  that is the explicit unwrapping of wrapped primitive values,  is unnecessary under Java 5 and newer, and can be safely removed.

Found with IntelliJ Analyze plugin.


mwithi (approved):
Yep, thanks!

PR #122, opened 2021-04-13
Chore: remove dead store

dbmalkovsky:
Remove dead stores (unneeded allocations and/or unused variable).

Found with SpotBugs.


mwithi (approved):
Thanks!

PR #129, opened 2021-05-10
Chore: remove unused class

dbmalkovsky:
Found no references to `OHFilterInvocationSecurityMetadataSource` in CORE, GUI, API, or UI repos; removing the class.


mwithi (approved):
And it seems not adding any value too... thanks!

PR #130, opened 2021-05-10
Chore: consistency

dbmalkovsky:
This was the only occurrence in the code where the `body` value was a variable and not `null` when there was no content.   It doesn't change the code but makes it consistent.

This was found with Spotbugs where it complained about loading a known null value.


mwithi (approved):
Thanks!

PR #131, opened 2021-05-10
Chore: use noneMatch() instead of negation

dbmalkovsky:
Use `noneMatch()` instead of negating `anyMatch()` which might be overlooked in a quick read.

Found with IntelliJ Analyze plugin.


mwithi (approved):
Alright, thanks!

PR #142, opened 2021-09-21
OP-468 update license headers

dbmalkovsky:
See OP-468

Additional files need update since last run of updating


mwithi (approved):
Thanks!

PR #151, opened 2021-12-28
OP-729 Update project to Java 8 and new time objects

dbmalkovsky:
See OP-729

Sorry but there are lots of changes; at least all the tests pass.

I also changed the jar version to 0.2 so as not to impact the UI development still using version 0.1


pom.xml
```diff
@@ -5,56 +5,47 @@
...
     <artifactId>openhospital-api</artifactId>
-    <version>0.0.1</version>
+    <version>0.0.2</version>
     <name>openhospital-api</name>
-    <description>Openhospital Spring Boot REST API </description>
+    <description>Open Hospital Spring Boot REST API</description>
 
     <properties>
-        <oh.version>1.11.0</oh.version>
+        <oh.version>1.11.1</oh.version>
```
mwithi, line 18:
```suggestion
        <oh.version>1.11.2</oh.version>
```

src/main/java/org/isf/disease/rest/DiseaseController.java
```diff
@@ -243,15 +243,15 @@ public DiseaseController(DiseaseBrowserManager diseaseManager, DiseaseMapper dis
...
 			throw new OHAPIException(new OHExceptionMessage(null, "duplicated disease description for the same disease type", OHSeverityLevel.ERROR), HttpStatus.INTERNAL_SERVER_ERROR);
 		}
-		
+
 		if (diseaseManager.newDisease(disease) != null) {
         	return ResponseEntity.status(HttpStatus.CREATED).body(diseaseDTO);
-        } else {
-        	throw new OHAPIException(new OHExceptionMessage(null, "disease is not created!", OHSeverityLevel.ERROR), HttpStatus.INTERNAL_SERVER_ERROR);
         }
+        throw new OHAPIException(new OHExceptionMessage(null, "disease is not created!", OHSeverityLevel.ERROR), HttpStatus.INTERNAL_SERVER_ERROR);
```
mwithi, line 254: maybe indentation?

pom.xml
dbmalkovsky: @mwithi I don't think I can do that as the main Core repo still has 1.11.1 in it.

src/main/java/org/isf/disease/rest/DiseaseController.java
dbmalkovsky: I reformatted the entire method.

pom.xml
mwithi: true... but I'm merging soon 1.11.2

mwithi (approved):
LGTM. Anyway I would ask a review to "ui" frontend to know if they can adopt this new J8 version

dbmalkovsky:
Because there is a jar version change they can take it anytime they want and is convenient for them.

mwithi:
> Because there is a jar version change they can take it anytime they want and is convenient for them.

yes, but if they can adopt it now with small changes on their side it is better to keep the full stack aligned IMHO and avoid the two components diverge too much

almeida1492 (approved):

mwithi:
Thanks a lot guys!

PR #159, opened 2022-04-01
updated pom.xml

AkshitKumar13:
i update the versions of parent and swagger and databases


src/main/java/org/isf/patient/rest/PatientController.java
```diff
@@ -1,9 +1,8 @@
 package org.isf.patient.rest;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
+import java.util.*;
```
dbmalkovsky: Project standards to not allow package (wildcard) imports; all imports must be enumerated.

src/main/java/org/isf/patient/rest/PatientController.java
```diff
@@ -105,40 +104,40 @@ public PatientController(PatientBrowserManager patientManager, PatientMapper pat
...
+            @RequestParam(value="secondName", defaultValue="", required = false) String secondName,
+            @RequestParam(value="birthDate", defaultValue="", required = false) String birthDate,
+            @RequestParam(value="address", defaultValue="", required = false) String address
     ) throws OHServiceException {
 
 
         List<Patient> patientList = null;
 
-        Map<String, String> params = new HashMap<String, String>();
+        Map<String, Object> params = new HashMap<String, Object>();
```
dbmalkovsky: Minimum JDK is JDK 8 which doesn't require data types for new statements; it gets them from the variable declaration.

dbmalkovsky:
It appears that the base line code you started with is not the latest version in develop; is that correct?

mwithi:
@AkshitKumar13 you didn't update this PR and time has past... setting back to draft before to close defintely.

Anyway, I suggest to sync your fork with [intesys/openhospital-api](https://github.com/intesys/openhospital-api) and to open pull-request against it (not against `informatici/openhospital-api`)

PR #171, opened 2022-05-02
OP-858 develop branch should use SNAPSHOT jar

dbmalkovsky:
See OP-858

This PR requires that the CORE PR is merged first: https://github.com/informatici/openhospital-core/pull/657

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,12 +10,12 @@
     </parent>
     <groupId>org.isf</groupId>
     <artifactId>openhospital-api</artifactId>
-    <version>0.0.2</version>
+    <version>1.12.0-SNAPSHOT</version>
     <name>openhospital-api</name>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -10,12 +10,12 @@
     </parent>
     <groupId>org.isf</groupId>
     <artifactId>openhospital-api</artifactId>
-    <version>0.0.2</version>
+    <version>1.12.0-SNAPSHOT</version>
```
mwithi, line 13: Not sure if we really need to change this... I think release 1.0 will be at the first release including this module (or anyway we cannot state now that it will with release 1.12)

dbmalkovsky: OK; reverted the change

mwithi: Alright! Thanks!

PR #200, opened 2023-01-24
OP-962 - session table for logged users

goto-eof:


src/main/java/org/isf/security/OHSimpleUrlAuthenticationSuccessHandler.java
```diff
@@ -65,14 +76,20 @@ public void onAuthenticationSuccess(
 		response.getWriter().append(mapper.writeValueAsString(loginResponse));
 		response.setStatus(200);
 
+		try {
+			this.httpSession.setAttribute("sessionAuditId",
+							sessionAuditManager.newSessionAudit(new SessionAudit(authentication.getName(), LocalDateTime.now(), null)));
+			System.out.println("AuditID: " + this.httpSession.getAttribute("sessionAuditId"));
```
dbmalkovsky, line 82: Why is there a `println`?

mwithi (approved):
LGTM. Thanks a lot!

PR #214, opened 2023-02-07
OH2-179 make get hospital endpoint public

ArnaudFonzam:
See OH2-179.

it is the new requirement in other to fix the name of the hospital in the login form

Paired with https://github.com/informatici/openhospital-ui/pull/456.


mwithi (dismissed):
As explained under the OH2-179 comments, if the intention is to SHOW the Hospital's name before the login, this change LGTM.

If the intention is to NOT show the Hospital's name, this change is not needed.

mwithi (commented):
As explained under the [OH2-179](https://openhospital.atlassian.net/browse/OH2-179) comments, if the intention is to SHOW the Hospital's name before the login, this change LGTM.

If the intention is to NOT show the Hospital's name, this change is not needed.

ArnaudFonzam:
but now i don't know if it is neccesary. because the first intention was to show the Hospital's name before the login

mwithi:
> but now i don't know if it is neccesary. because the first intention was to show the Hospital's name before the login

not exactly... the initial intention was to replace the placeholder "Plainsboro Hospital" with the real name. Since this placeholder was on the login page, you asked to make the /hospitals endpoint public.

But in https://github.com/informatici/openhospital-ui/pull/456 you put just "Sign in" before the login, hence my question.

mwithi (approved):
Since https://github.com/informatici/openhospital-ui/pull/456 has been merged we can merge also this.

PR #233, opened 2023-03-31
OH2-188 Add parameter status on api getlaboratoryForPrint

ArnaudFonzam:


src/main/java/org/isf/lab/rest/LaboratoryController.java
```diff
@@ -540,7 +541,13 @@ public ResponseEntity<List<LabWithRowsDTO>> getLaboratoryForPrint(
 						HttpStatus.INTERNAL_SERVER_ERROR);
 			}
 		}
-		List<Laboratory> laboratoryList = laboratoryManager.getLaboratory(examName, dateF, dateT, patient);
+		List<Laboratory> laboratoryList = new ArrayList<Laboratory>();
```
dbmalkovsky, line 544:
```suggestion
		List<Laboratory> laboratoryList = new ArrayList<>();
```
Currrent Java language doesn't require the type to be defined on both sides

PR #240, opened 2023-04-06
OP-1066 Use new shorthand OHExceptionMessage() constructor for packages A and D

dbmalkovsky:
See OP-1066

Changes for packages starting with 'A' and 'D'

Some additional message text cleanup too.


mwithi (approved):
LGTM. Thanks a lot!

PR #254, opened 2023-04-19
Chore: string comparisons should use equals() and not ==

dbmalkovsky:


dbmalkovsky:
Causes test to fail; will need more debugging later.

PR #260, opened 2023-04-22
Chore: fix mismatch between Javadoc and code; fix grammar too

dbmalkovsky:


src/main/java/org/isf/vaccine/rest/VaccineController.java
```diff
@@ -166,7 +166,7 @@ public ResponseEntity<Boolean> deleteVaccine(@PathVariable("code") String code)
      * Check if the code is already used by other vaccine.
      *
      * @param code
-     * @return true if it is already use, false otherwise
+     * @return {@code true}if it is already used, false otherwise
```
mwithi, line 169:
```suggestion
     * @return {@code true} if it is already used, false otherwise
```

mwithi (approved):
LGTM. Thanks a lot!

PR #262, opened 2023-04-25
OH2-182 Standardize error message

ArnaudFonzam:


bundle/language_en.properties
```diff
@@ -0,0 +1,173 @@
...
+stats.filenotfound = File not found.
+stats.failedtoloadthefiletype = Failed to load the files type.
+supplier.notfound = Supplier not found.
+supplier.isnotcreated = Supplier is not created.
+supplier.isnotupdated = Supplier is not updated.
+therapy.arenotreplaced = Therapies are not replaced.
+therapy.arenotdeleted = Therapies are not deleted.
+vaccine.isnotcreated = Vaccine is not created.
+vaccine.isnotupdated = Vaccine is not updated.
+vaccine.isnotdeleted = vaccine is not updated.
```
dbmalkovsky, line 160: There are so many things that need to be changed in this PR but I'll  only note one of them here.  The key is includes the text `deleted` but the property says `updated`

bundle/language_en.properties
```diff
@@ -0,0 +1,175 @@
...
+supplier.notfound = Supplier not found.
+supplier.isnotcreated = Supplier is not created.
+supplier.isnotupdated = Supplier is not updated.
+sql.problemsoccurredwithserverconnection = Problems occurred with server connection.
+therapy.arenotreplaced = Therapies are not replaced.
+therapy.arenotdeleted = Therapies are not deleted.
+vaccine.isnotcreated = Vaccine is not created.
+vaccine.isnotupdated = Vaccine is not updated.
+vaccine.isnotdeleted = vaccine is not deleted.
+vaccinetype.isnotcreated = vaccine type is not created.
```
dbmalkovsky, line 163: Why don't all message text start with a capital letter?

src/main/java/org/isf/shared/FormatErrorMessage.java
```diff
@@ -0,0 +1,16 @@
+package org.isf.shared;
```
dbmalkovsky, line 1: Missing license header

bundle/language_en.properties
ArnaudFonzam: I will update its. tky

src/main/java/org/isf/shared/FormatErrorMessage.java
```diff
@@ -0,0 +1,16 @@
...
+
+public class FormatErrorMessage {
```
dbmalkovsky, line 3: Why is this even necessary?  You are just working against the standards that exist for the project for a long time and implementing new "standards" .  This just seems wrong and unnecessary.

dbmalkovsky: My point is (and maybe most people would disagree with me) that it seems confusing that two different "standards" are used in the same project.  Why did the front-end feel like they needed a different "standard format"?   There are lots of "standards" in the project that I'd like to change but don't because it would make things difficult for anyone who works on multiple parts of the project.

ArnaudFonzam: Okay i think it is better to discuss with @mwithi to have the good way to do because the work that i made was to analyse and give a solution of the problem.

ArnaudFonzam: Okay

PR #264, opened 2023-05-02
OP-1067: Change the default hardcoded value of password length for the one specified in the settings

Verdoso:
On an unrelated note, accessing the settings through static variables is kind of ugly and it would look nicer to do it through @Configuration classes. It would make testing easier also, as static values are harder to mock.


dbmalkovsky (approved):
LGTM

PR #265, opened 2023-05-04
OP-1077 Remove password length check as password handling is done in Core

dbmalkovsky:
See OP-1077


mwithi (approved):
LGTM. Thanks a lot!

PR #267, opened 2023-05-08
OP-1083: changes from removing deprecated method from PatientBrowserManager

dbmalkovsky:
See OP-1083

Goes along with https://github.com/informatici/openhospital-core/pull/976

### DIFF
--- a/src/main/java/org/isf/accounting/rest/BillController.java
+++ b/src/main/java/org/isf/accounting/rest/BillController.java
@@ -111,7 +111,7 @@ public class BillController {
 
 		Bill bill = billMapper.map2Model(newBillDto.getBill());
 
-		Patient pat = patientManager.getPatientByName(bill.getPatName()); // FIXME: verify why we were searching by name
+		Patient pat = patientManager.getPatientById(bill.getBillPatient().getCode());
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/isf/accounting/rest/BillController.java
```diff
@@ -111,7 +111,7 @@ public ResponseEntity<FullBillDTO> newBill(@RequestBody FullBillDTO newBillDto)
 
 		Bill bill = billMapper.map2Model(newBillDto.getBill());
 
-		Patient pat = patientManager.getPatientByName(bill.getPatName()); // FIXME: verify why we were searching by name
+		Patient pat = patientManager.getPatientById(bill.getBillPatient().getCode());
```
mwithi, line 114: @ArnaudFonzam can you please clarify why we were searching by name instead of by id? Can we safely remove this?

ArnaudFonzam: @mwithi  yes we can change and it is  on controller and the ui is not available yet. it is better to search by id.

mwithi: Ok, thanks!

PR #268, opened 2023-05-09
OP-1086: removed deprecated methods from org.isf.lab.model.Laboratory

dbmalkovsky:
See OP-1086

Using new Core jar; rebuilt and reran all tests successfully

Dependent on: https://github.com/informatici/openhospital-core/pull/978


mwithi (approved):
LGTM. Thanks a lot!

PR #271, opened 2023-05-11
OP-1086: removed deprecated methods from org.isf.medicalstockward.manager,MovWardBrowserManager.

dbmalkovsky:
See OP-1090

Removed methods from Core, rebuilt the jar, ran the unit tests.  All was fine.
See https://github.com/informatici/openhospital-core/pull/983

Used new jar and rebuilt GUI and there were no issues.

Used new jar and rebuilt API and there were errors.   I commented out three methods that were causing problems.   I checked the UI code base and it appears that these "/medicalstockward" endpoints are not being used but only defined.   I left them in the API code for easier updating of the API and UI code bases but those more familiar with the UI.

### DIFF
--- a/src/main/java/org/isf/medicalstockward/rest/MedicalStockWardController.java
+++ b/src/main/java/org/isf/medicalstockward/rest/MedicalStockWardController.java
@@ -122,21 +122,21 @@ public class MedicalStockWardController {
 		return ResponseEntity.ok(movWardBrowserManager.getCurrentQuantityInWard(wards.get(0), medical));
 	}
 
-	/**
-	 * Gets all the {@link MovementWard}s.
-	 *
-	 * @return all the retrieved movements ward.
-	 * @throws OHServiceException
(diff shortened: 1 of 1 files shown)


dbmalkovsky:
They wouldn't compile; the basic problem is that `movWardBrowserManager.getMovementWard()` no longer exists.  Given that the code isn't being used this appeared to be the safest route until it is determined how and when the code is needed.

mwithi:
> They wouldn't compile; the basic problem is that `movWardBrowserManager.getMovementWard()` no longer exists. Given that the code isn't being used this appeared to be the safest route until it is determined how and when the code is needed.

Ok, but I would add in the comment the reason why we are keeping them... something like:
`// TODO: needs update`

dbmalkovsky:
OK; done.

mwithi (approved):
LGTM. Thanks a lot!

PR #272, opened 2023-05-12
OH2-199: Fix laboratory api returning null for labDate and registrationDate

ArnaudFonzam:
See OH2-199


mwithi (approved):
LGTM. Thanks a lot!

PR #277, opened 2023-05-23
Chore: Spring Security clean up

mwithi:
PR just for sharing and commenting


src/main/java/org/isf/config/SecurityConfig.java
```diff
@@ -108,102 +108,169 @@ public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
...
+						.antMatchers(HttpMethod.PUT, "/diseasetypes/**").hasAuthority("admin")
+						.antMatchers(HttpMethod.DELETE, "/diseasetypes/**").hasAuthority("admin")
 						.antMatchers(HttpMethod.GET, "/diseasetypes/**").hasAnyAuthority("admin", "guest")
 						// opd
-						.antMatchers(HttpMethod.POST, "/opds/**").permitAll().antMatchers(HttpMethod.PUT, "/opds/**").permitAll()
-						.antMatchers(HttpMethod.DELETE, "/opds/**").permitAll().antMatchers(HttpMethod.PATCH, "/opds/**").permitAll()
+						.antMatchers(HttpMethod.POST, "/opds/**").permitAll()
+						.antMatchers(HttpMethod.PUT, "/opds/**").permitAll()
+						.antMatchers(HttpMethod.DELETE, "/opds/**").permitAll()
+						.antMatchers(HttpMethod.PATCH, "/opds/**").permitAll()
```
mwithi, line 230: Hi @ArnaudFonzam, any reason why from this point they are all `permitAll()` except `GET` methods?

ArnaudFonzam: Hello @mwithi no we can also fix permitAll() for Get methods.

mwithi: what do you mean? GET got always (correctly) `hasAnyAuthority("admin", "guest")`
I mean that only others need to get `hasAnyAuthority("admin")`

After of course we should start to manage roles in the DB and in the code

ArnaudFonzam: Okay for me not problem we can remove guest and do it like you suggest.

mwithi: Guest user can remain, the point is to apply these changes that should be consistent now, please review this PR. Thanks!

PR #303, opened 2023-07-29
Chore: set method should return void

dbmalkovsky:
Clean rebuild and all the tests run after the change


mwithi (approved):
LGTM. Thanks a lot!

PR #304, opened 2023-07-29
Chore: remove file that is just one large comment; it is unused

dbmalkovsky:
Even the license header and the class name are commented out.....


mwithi (approved):
LGTM. Thanks a lot!

PR #306, opened 2023-07-29
Chore: remove deprecated variable and associated methods

dbmalkovsky:
Ran `mvn clean test` without error after removing the variable and the associated methods.


mwithi (approved):
LGTM. Thanks a lot!

PR #307, opened 2023-07-29
Chore: use isEmpty() instead of testing for size of the collection

dbmalkovsky:
Changing these 3 occurrences of `size() > 0` matches the other 153 occurrences in the project


mwithi (approved):
LGTM. Thanks a lot!

PR #310, opened 2023-08-03
OH2-227: Persist the opd ward that comes from the front end

ArnaudFonzam:
See OH2-227.
Paired with https://github.com/informatici/openhospital-core/pull/1059.


mwithi (approved):
LGTM. Thanks a lot!

PR #317, opened 2023-08-05
Chore: use `isEmpty()` like every other class in the project

dbmalkovsky:


mwithi (changes requested):
LGTM. Thanks a lot!

mwithi (approved):
LGTM. Thanks a lot!

PR #319, opened 2023-08-06
Chore: remove unused class

dbmalkovsky:


mwithi (approved):
This class is a WIP, but I agree it can be removed and added only when needed

PR #322, opened 2023-08-15
[OP-1121] Correct the jar version into dockerfile.backend

Tudor44:


README.md
```diff
@@ -90,10 +90,10 @@ You can see Swagger Api Documentation at: http://localhost:8080/swagger-ui/
 
 Make sure you have docker with docker-compose installed, then run the following commands:
 
-- copy `dotenv` file into `.env` and set variables as needed (the SHA-256 jwt token is needed)
+- rename the `dotenv` file in `.env` and set variables as needed (the SHA-256 jwt token is needed)
```
mwithi, line 93: renaming would result in a different git status and may be commited by mistake

Tudor44: > renaming would result in a different git status and may be commited by mistake

Yes, I would say rename .dotenv, but keeping the original .dotenv versioned as is into current branch. 
It should be useful a .gitignore entry for .env

mwithi: oh, I supposed was already there in .gitignore like did in other repos, where we also use the same instructions: dotenv and copy to .env

could you please revert this line in README and add .env to .gitignore?

Tudor44: > oh, I supposed was already there in .gitignore like did in other repos, where we also use the same instructions: dotenv and copy to .env
> 
> could you please revert this line in README and add .env to .gitignore?

Ok sure, I ll add it in develop only?

mwithi:
was this PR meant to be closed?

README.md
Tudor44: Added new PR with JIRA assocations.

mwithi:
> was this PR meant to be closed?

ok, sorry, I've seen the new PR now

Tudor44:
> was this PR meant to be closed?

Hi @mwithi , sorry I added it in a new PR

PR #327, opened 2023-08-23
Chore: single call instead of new() followed by addAll()

dbmalkovsky:
Use a single call to a parameterized constructor which simplifies the code.   For some collection types the replacement is more performant.


mwithi (approved):
LGTM. Thanks a lot!

PR #328, opened 2023-08-23
Chore: use char when String is a single character

dbmalkovsky:
Interestingly it had been a single character for a long time until some mass mergers were undertaken.  Using a single character is usually more performant.


mwithi (approved):
LGTM. Thanks a lot!

PR #336, opened 2023-09-27
OP-1130 improve JPA methods

dbmalkovsky:
See OP-1130

This is for the accounting package and is matched with the CORE PR: https://github.com/informatici/openhospital-core/pull/1102


mwithi (approved):
LGTM. Thanks a lot!

PR #338, opened 2023-09-28
OP-1130 improve JPA methods - admission package

dbmalkovsky:
See OP-1130

Changes for the admission package

Corresponds the CORE repo changes in https://github.com/informatici/openhospital-core/pull/1104

### DIFF
--- a/src/main/java/org/isf/admission/rest/AdmissionController.java
+++ b/src/main/java/org/isf/admission/rest/AdmissionController.java
@@ -332,15 +332,12 @@ public class AdmissionController {
 	@DeleteMapping(value = "/admissions/{id}", produces = MediaType.APPLICATION_JSON_VALUE)
 	public ResponseEntity<Boolean> deleteAdmissionType(@PathVariable("id") int id) throws OHServiceException {
 		LOGGER.info("setting admission to deleted: {}", id);
-		boolean isDeleted;
 		Admission admission = admissionManager.getAdmission(id);
-		if (admission != null) {
-			isDeleted = admissionManager.setDeleted(id);
-		} else {
+		if (admission == null) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/isf/admission/rest/AdmissionController.java
```diff
@@ -332,15 +332,12 @@ public ResponseEntity<Integer> getUsedWardBed(@RequestParam("wardid") String war
...
-		if (admission != null) {
-			isDeleted = admissionManager.setDeleted(id);
-		} else {
+		if (admission == null) {
 			return ResponseEntity.status(HttpStatus.NOT_FOUND).body(null);
 		}
-
-		return ResponseEntity.ok(isDeleted);
+		admissionManager.setDeleted(id);
+		return ResponseEntity.ok(true);
```
mwithi, line 340: what if the `setDeleted(id)` returns null?

mwithi: Up

dbmalkovsky: In this method it will never return `null` as lines 335-337 are:
<pre>
                Admission admission = admissionManager.getAdmission(id);
		if (admission == null) {
			return ResponseEntity.status(HttpStatus.NOT_FOUND).body(null);
		}
</pre>
Which means if the admission with that `id` doesn't exist it never executes `admissionManager.setDeleted(id)` and returns`NOT_FOUND`.  The only way `setDeleted(id)` returns `null` is if the record doesn't exist but it is short circuited by the  `null` check in this file.

mwithi: sorry, I made an oversight

mwithi (approved):
LGTM. Thanks a lot!

PR #341, opened 2023-10-03
OP-1130 JPA admtype changes (take 2)

dbmalkovsky:
See OP-1130

See Core PR: https://github.com/informatici/openhospital-core/pull/1111 for more information.


dbmalkovsky:
The approach will not be used; closing.

PR #357, opened 2023-10-09
OP-1130 improve JPA methods - exatype package

dbmalkovsky:
See OP-1130

Changes for exatype package

Matches changes in CORE PR https://github.com/informatici/openhospital-core/pull/1119

### DIFF
--- a/src/main/java/org/isf/exatype/rest/ExamTypeController.java
+++ b/src/main/java/org/isf/exatype/rest/ExamTypeController.java
@@ -71,10 +71,6 @@ public class ExamTypeController {
 
 		ExamType examType = examTypeMapper.map2Model(newExamType);
 		ExamType createdExamType = examTypeBrowserManager.newExamType(examType);
-		if (!examTypeBrowserManager.isCodePresent(examType.getCode())) {
-			throw new OHAPIException(new OHExceptionMessage("Exam Type not created."));
-		}
-
 		return ResponseEntity.status(HttpStatus.CREATED).body(examTypeMapper.map2DTO(createdExamType));
 	}
 
@@ -106,7 +102,7 @@ public class ExamTypeController {
 
     @DeleteMapping(value = "/examtypes/{code:.+}", produces = MediaType.APPLICATION_JSON_VALUE)
     public ResponseEntity<Boolean> deleteExamType(@PathVariable String code) throws OHServiceException {
-	    LOGGER.info("Delete exam type code: {}", code);
+	    LOGGER.info("Delete Exam Type code: {}", code);
         Optional<ExamType> examType = examTypeBrowserManager.getExamType().stream().filter(e -> e.getCode().equals(code)).findFirst();
         if (examType.isEmpty()) {
             throw new OHAPIException(new OHExceptionMessage("Exam Type not found."));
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/exatype/rest/ExamTypeController.java
```diff
@@ -71,8 +71,8 @@ public ResponseEntity<ExamTypeDTO> newExamType(@RequestBody ExamTypeDTO newExamT
 
 		ExamType examType = examTypeMapper.map2Model(newExamType);
 		ExamType createdExamType = examTypeBrowserManager.newExamType(examType);
-		if (createdExamType == null) {
-			throw new OHAPIException(new OHExceptionMessage("Exam Type type not created."));
+		if (!examTypeBrowserManager.isCodePresent(examType.getCode())) {
+			throw new OHAPIException(new OHExceptionMessage("Exam Type not created."));
```
mwithi, line 75: Why we check the existance of `createdExamType` with a new query on the DB rather than just check if the returning object is `null`?

src/main/java/org/isf/exatype/rest/ExamTypeController.java
```diff
@@ -89,12 +89,8 @@ public ResponseEntity<ExamTypeDTO> updateExamType(@PathVariable String code, @Re
...
 
         ExamType examType = examTypeMapper.map2Model(updateExamType);
-        ExamType exType = examTypeBrowserManager.updateExamType(examType);
-        if (exType == null) {
-            throw new OHAPIException(new OHExceptionMessage("Exam Type not updated."));
-        }
-
-        return ResponseEntity.ok(examTypeMapper.map2DTO(exType));
+        ExamType updatedExamType = examTypeBrowserManager.updateExamType(examType);
+        return ResponseEntity.ok(examTypeMapper.map2DTO(updatedExamType));
```
mwithi, line 93: what if `updatedExamType` is null?

src/main/java/org/isf/exatype/rest/ExamTypeController.java
dbmalkovsky: Because the contract for the underlying `save` method in the CRUD repository says:
<pre>
 @return the saved entity; will never be {@literal null}
</pre>

src/main/java/org/isf/exatype/rest/ExamTypeController.java
dbmalkovsky: same comment as above

src/main/java/org/isf/exatype/rest/ExamTypeController.java
mwithi: Ok, right. But so maybe we don't need this check anymore? The entity is either saved (and returned) or an exception is thrown (and catched)?

src/main/java/org/isf/exatype/rest/ExamTypeController.java
mwithi: yeah, so maybe above we don't need a check like here?

src/main/java/org/isf/exatype/rest/ExamTypeController.java
mwithi: Ok, I will test the catch and see what to do in another issue probably. Thanks!

mwithi (approved):
LGTM. Thanks a lot!

PR #371, opened 2023-10-23
OH2-241 | Leverage examination.properties settings in ExaminationController

ArnaudFonzam:
- update examination.properties
- create validation method to validate the Examination properties on ExaminationController
- remove enum Bowel, Diurese, Ausculation to core


src/main/java/org/isf/examination/rest/ExaminationController.java
```diff
@@ -268,4 +226,42 @@ public ResponseEntity<List<PatientExaminationDTO>> getByPatientId(@PathVariable
 			return ResponseEntity.ok(listPatientExaminationDTO);
 		}
 	}
+	
+	public void validateExamination(PatientExaminationDTO newPatientExamination) throws OHServiceException {
+		ExaminationParameters.initialize();
```
dbmalkovsky, line 231: Formatting of these lines look incorrect

rsc/examination.properties
```diff
@@ -24,7 +24,7 @@ TEMP_STEP = 0.1
 #SAT_UNIT = %
 SAT_INIT = 98
 SAT_MIN = 50
-#SAT_MAX = 100
+SAT_MAX = 100
```
mwithi, line 27: This was commented because the Saturation cannot be above 100%, so it is fine to have it hardcoded

src/main/java/org/isf/examination/rest/ExaminationController.java
```diff
@@ -81,38 +85,15 @@ public ResponseEntity<Boolean> newPatientExamination(@RequestBody PatientExamina
...
-        }
-        if (newPatientExamination.getPex_rr() < 0 || newPatientExamination.getPex_rr() > 100) {
-        	throw new OHAPIException(new OHExceptionMessage("Respiratory rate should be between 0 and 100."));
+        if (newPatientExamination.getPex_auscultation() == null) {
+        	newPatientExamination.setPex_auscultation(Ausculation.normal);
         }
-        if (newPatientExamination.getPex_diuresis() < 0 || newPatientExamination.getPex_diuresis() > 2500) {
-        	throw new OHAPIException(new OHExceptionMessage("Diuresis should be between 0 and 2500."));
+        if (newPatientExamination.getPex_bowel_desc() == null) {
+        	newPatientExamination.setPex_bowel_desc(Bowel.regular);
```
mwithi, line 96: this is not correct, every parameter could be null, except Height and Weight that are compulsory.

The check here is to verify that the passed string is within the allowed ones (which are mapped and hardcoded)

Why we don't leverage the validation in core?

src/main/java/org/isf/examination/rest/ExaminationController.java
```diff
@@ -81,38 +85,15 @@ public ResponseEntity<Boolean> newPatientExamination(@RequestBody PatientExamina
...
+        if (newPatientExamination.getPex_diuresis_desc() == null) {
+        	newPatientExamination.setPex_diuresis_desc(Diurese.physiological);
         }
-        if (newPatientExamination.getPex_hgt() < 30 || newPatientExamination.getPex_temp() > 600.0) {
-        	throw new OHAPIException(new OHExceptionMessage("HGT should be between 30 and 600."));
-        }
-        if (newPatientExamination.getPex_rr() < 0 || newPatientExamination.getPex_rr() > 100) {
-        	throw new OHAPIException(new OHExceptionMessage("Respiratory rate should be between 0 and 100."));
+        if (newPatientExamination.getPex_auscultation() == null) {
+        	newPatientExamination.setPex_auscultation(Ausculation.normal);
```
mwithi, line 93: this is not correct, every parameter could be null, except Height and Weight that are compulsory.

The check here is to verify that the passed string is within the allowed ones (which are mapped and hardcoded)

Why we don't leverage the validation in core?

src/main/java/org/isf/examination/rest/ExaminationController.java
```diff
@@ -81,38 +85,15 @@ public ResponseEntity<Boolean> newPatientExamination(@RequestBody PatientExamina
...
-        	throw new OHAPIException(new OHExceptionMessage("Heart rate should be between 0 and 240."));
-        }
-        if (newPatientExamination.getPex_temp() < 30.0 || newPatientExamination.getPex_temp() > 50.0) {
-        	throw new OHAPIException(new OHExceptionMessage("The temperature should be between 30 and 50."));
-        }
-        if (newPatientExamination.getPex_sat() < 50.0 || newPatientExamination.getPex_temp() > 100.0) {
-        	throw new OHAPIException(new OHExceptionMessage("The saturation should be between 50 and 100."));
+        validateExamination(newPatientExamination);
+        if (newPatientExamination.getPex_diuresis_desc() == null) {
+        	newPatientExamination.setPex_diuresis_desc(Diurese.physiological);
```
mwithi, line 90: this is not correct, every parameter could be null, except Height and Weight that are compulsory.

The check here is to verify that the passed string is within the allowed ones (which are mapped and hardcoded)

Why we don't leverage the validation in core?

src/main/java/org/isf/examination/rest/ExaminationController.java
```diff
@@ -268,4 +226,42 @@ public ResponseEntity<List<PatientExaminationDTO>> getByPatientId(@PathVariable
...
+	if (newPatientExamination.getPex_height() < ExaminationParameters.HEIGHT_MIN || newPatientExamination.getPex_height() > ExaminationParameters.HEIGHT_MAX) {
+        	throw new OHAPIException(new OHExceptionMessage("The size should be between 0 and 250."));
+        }
+        if (newPatientExamination.getPex_weight() < ExaminationParameters.WEIGHT_MIN || newPatientExamination.getPex_weight() > ExaminationParameters.WEIGHT_MAX) {
+        	throw new OHAPIException(new OHExceptionMessage("The weight should be between 0 and 200."));
+        }
+        if (newPatientExamination.getPex_ap_min() < ExaminationParameters.AP_MIN_INIT) {
+        	throw new OHAPIException(new OHExceptionMessage("The minimum blood pressure must be at least 80."));
+        }
+        if (newPatientExamination.getPex_ap_min() > newPatientExamination.getPex_ap_max() ) {
```
mwithi, line 242: The error message should include the same MIN and MAX (or INIT) values.

Why we don't leverage the validation in core?

mwithi (changes requested):
Validation is already done in core.
Values should not be forced to their "normal" values, because they may have not been observed at all. There is a difference between "normal observation" and "not observed"

src/main/java/org/isf/examination/rest/ExaminationController.java
ArnaudFonzam: that is because in ExaminationParameters I have see the default value for those param

src/main/java/org/isf/examination/rest/ExaminationController.java
ArnaudFonzam: I think it is because the error message return by this method is only use on the api if I add this methode on the core it will implies to update also the bundle file of gui and I don't have access to that file.

src/main/java/org/isf/examination/rest/ExaminationController.java
mwithi: they are not "default" but "init" values, in this case means "normal" values also, but if you don't ask or check on the patient that parameter you cannot assume is "normal", you just don't know it.

Indeed, in the GUI, there is a checkbox in order to enter or not one parameter (except for height and weight): if you want to enter it, you enable it and it starts with the "init" value, then you change it.

I don't know how is the best way to achieve this in the UI, but the API should allow `null` values (except for height and weight)

src/main/java/org/isf/examination/rest/ExaminationController.java
mwithi: yes you are right, we should totally avoid the use of bundles in the core... I will study how to do it.

as for now, please make `OHExceptionMessages` to include the parameters used in the comparations.

src/main/java/org/isf/examination/rest/ExaminationController.java
```diff
@@ -256,7 +196,7 @@ public ResponseEntity<Page<PatientExaminationDTO>> getLastNByPatID(@RequestParam
 	}
 
 	@GetMapping(value = "/examinations/byPatientId/{patId}", produces = MediaType.APPLICATION_JSON_VALUE)
-	public ResponseEntity<List<PatientExaminationDTO>> getByPatientId(@PathVariable Integer patId) throws OHServiceException {
+    public ResponseEntity<List<PatientExaminationDTO>> getByPatientId(@PathVariable Integer patId) throws OHServiceException {
```
mwithi, line 199: please fix indentation

src/main/java/org/isf/examination/rest/ExaminationController.java
```diff
@@ -268,4 +208,42 @@ public ResponseEntity<List<PatientExaminationDTO>> getByPatientId(@PathVariable
...
+        }
+        if (newPatientExamination.getPex_weight() < ExaminationParameters.WEIGHT_MIN || newPatientExamination.getPex_weight() > ExaminationParameters.WEIGHT_MAX) {
+        	throw new OHAPIException(new OHExceptionMessage("The weight should be between"+ExaminationParameters.WEIGHT_MIN+" and "+ExaminationParameters.WEIGHT_MAX));
+        }
+        if (newPatientExamination.getPex_ap_min() < ExaminationParameters.AP_MIN_INIT) {
+        	throw new OHAPIException(new OHExceptionMessage("The minimum blood pressure must be at least "+ExaminationParameters.AP_MIN_INIT));
+        }
+        if (newPatientExamination.getPex_ap_min() > newPatientExamination.getPex_ap_max() ) {
+        	throw new OHAPIException(new OHExceptionMessage("The minimum blood pressure must be lower than "+ExaminationParameters.AP_MAX_INIT));
+        }
```
mwithi, line 226: please fix indentation

src/main/java/org/isf/examination/rest/ExaminationController.java
```diff
@@ -268,4 +208,42 @@ public ResponseEntity<List<PatientExaminationDTO>> getByPatientId(@PathVariable
...
+		
+	if (newPatientExamination.getPex_height() < ExaminationParameters.HEIGHT_MIN || newPatientExamination.getPex_height() > ExaminationParameters.HEIGHT_MAX) {
+        	throw new OHAPIException(new OHExceptionMessage("The size should be between "+ExaminationParameters.HEIGHT_MIN+" and "+ExaminationParameters.HEIGHT_MAX));
+        }
+        if (newPatientExamination.getPex_weight() < ExaminationParameters.WEIGHT_MIN || newPatientExamination.getPex_weight() > ExaminationParameters.WEIGHT_MAX) {
+        	throw new OHAPIException(new OHExceptionMessage("The weight should be between"+ExaminationParameters.WEIGHT_MIN+" and "+ExaminationParameters.WEIGHT_MAX));
+        }
+        if (newPatientExamination.getPex_ap_min() < ExaminationParameters.AP_MIN_INIT) {
+        	throw new OHAPIException(new OHExceptionMessage("The minimum blood pressure must be at least "+ExaminationParameters.AP_MIN_INIT));
+        }
```
mwithi, line 223: What if the `AP_MIN` is not passed at all (means not-observed)?

Anyway:
```suggestion
        if (newPatientExamination.getPex_ap_min() < ExaminationParameters.AP_MIN_INIT) {
        	throw new OHAPIException(new OHExceptionMessage("The minimum blood pressure must be at least " + ExaminationParameters.AP_MIN_INIT));
        }
```

src/main/java/org/isf/examination/rest/ExaminationController.java
mwithi, line 226: What if the `AP_MIN` and `AP_MAX` are not passed at all (means not-observed)?

Anyway:
```suggestion
        if (newPatientExamination.getPex_ap_min() > newPatientExamination.getPex_ap_max() ) {
        	throw new OHAPIException(new OHExceptionMessage("The minimum blood pressure must be lower than " + ExaminationParameters.AP_MAX_INIT));
        }
```

src/main/java/org/isf/examination/rest/ExaminationController.java
mwithi, line 223: the AP_MIN_INIT value is just the initial value for the FE, it is not the MIN value to be checked, remove this check

```suggestion
```

src/main/java/org/isf/examination/rest/ExaminationController.java
```diff
@@ -268,4 +208,42 @@ public ResponseEntity<List<PatientExaminationDTO>> getByPatientId(@PathVariable
...
+        if (newPatientExamination.getPex_ap_min() > newPatientExamination.getPex_ap_max() ) {
+        	throw new OHAPIException(new OHExceptionMessage("The minimum blood pressure must be lower than "+ExaminationParameters.AP_MAX_INIT));
+        }
+        if (newPatientExamination.getPex_ap_max() > ExaminationParameters.AP_MAX_INIT) {
+        	throw new OHAPIException(new OHExceptionMessage("The maximum blood pressure must be lower than "+ExaminationParameters.AP_MAX_INIT));
+        }
```
mwithi, line 229: the AP_MAX_INIT value is just the initial value for the FE, it is not the MAX value to be checked, remove this check
```suggestion
```

src/main/java/org/isf/examination/rest/ExaminationController.java
mwithi, line 226:
```suggestion
        if (newPatientExamination.getPex_ap_min() > newPatientExamination.getPex_ap_max() ) {
        	throw new OHAPIException(new OHExceptionMessage("The minimum blood pressure must be lower than the maximum blood pressure"));
        }
```

src/main/java/org/isf/examination/rest/ExaminationController.java
mwithi, line 1: In general, except height and weight that are always compulsory, the other parameters should be checked only if not null

src/main/java/org/isf/examination/rest/ExaminationController.java
```diff
@@ -268,4 +208,42 @@ public ResponseEntity<List<PatientExaminationDTO>> getByPatientId(@PathVariable
...
+	
+    public void validateExamination(PatientExaminationDTO newPatientExamination) throws OHServiceException {
+	ExaminationParameters.initialize();
+		
+	if (newPatientExamination.getPex_height() < ExaminationParameters.HEIGHT_MIN || newPatientExamination.getPex_height() > ExaminationParameters.HEIGHT_MAX) {
+        	throw new OHAPIException(new OHExceptionMessage("The size should be between "+ExaminationParameters.HEIGHT_MIN+" and "+ExaminationParameters.HEIGHT_MAX));
+        }
+        if (newPatientExamination.getPex_weight() < ExaminationParameters.WEIGHT_MIN || newPatientExamination.getPex_weight() > ExaminationParameters.WEIGHT_MAX) {
+        	throw new OHAPIException(new OHExceptionMessage("The weight should be between"+ExaminationParameters.WEIGHT_MIN+" and "+ExaminationParameters.WEIGHT_MAX));
+        }
```
mwithi, line 220: what if `null` values are passed?
Anyway:
```suggestion
	if (newPatientExamination.getPex_height() < ExaminationParameters.HEIGHT_MIN || newPatientExamination.getPex_height() > ExaminationParameters.HEIGHT_MAX) {
        	throw new OHAPIException(new OHExceptionMessage("The size should be between "+ExaminationParameters.HEIGHT_MIN+" and " + ExaminationParameters.HEIGHT_MAX));
        }
        if (newPatientExamination.getPex_weight() < ExaminationParameters.WEIGHT_MIN || newPatientExamination.getPex_weight() > ExaminationParameters.WEIGHT_MAX) {
        	throw new OHAPIException(new OHExceptionMessage("The weight should be between"+ExaminationParameters.WEIGHT_MIN+" and " + ExaminationParameters.WEIGHT_MAX));
        }
```

src/main/java/org/isf/examination/rest/ExaminationController.java
```diff
@@ -268,4 +208,42 @@ public ResponseEntity<List<PatientExaminationDTO>> getByPatientId(@PathVariable
...
+        if (newPatientExamination.getPex_hgt() < ExaminationParameters.HGT_MIN || newPatientExamination.getPex_temp() > ExaminationParameters.HGT_MAX) {
+        	throw new OHAPIException(new OHExceptionMessage("HGT should be between "+ExaminationParameters.HGT_MIN+" and "+ExaminationParameters.HGT_MAX));
+        }
+        if (newPatientExamination.getPex_rr() < ExaminationParameters.RR_MIN || newPatientExamination.getPex_rr() > ExaminationParameters.RR_MAX) {
+        	throw new OHAPIException(new OHExceptionMessage("Respiratory rate should be between "+ExaminationParameters.RR_MIN+" and "+ExaminationParameters.RR_MAX));
+        }
+        if (newPatientExamination.getPex_diuresis() < ExaminationParameters.DIURESIS_MIN || newPatientExamination.getPex_diuresis() > ExaminationParameters.DIURESIS_MAX) {
+        	throw new OHAPIException(new OHExceptionMessage("Diuresis should be between "+ExaminationParameters.DIURESIS_MIN+" and "+ExaminationParameters.DIURESIS_MAX));
+        }
```
mwithi, line 247: In general, except height and weight that are always compulsory, the other parameters should be checked only if not null (means not observed)

mwithi (changes requested):
Change requests, please check comments

src/main/java/org/isf/examination/rest/ExaminationController.java
ArnaudFonzam: okay

src/main/java/org/isf/examination/rest/ExaminationController.java
```diff
@@ -24,10 +24,14 @@
 import java.util.List;
 import java.util.stream.Collectors;
 
+import org.isf.examination.dto.Ausculation;
+import org.isf.examination.dto.Bowel;
+import org.isf.examination.dto.Diurese;
```
mwithi, line 29: These seem not used, please organize the imports

src/main/java/org/isf/examination/rest/ExaminationController.java
mwithi, line 1: Alos remove these lines from both newPatientExamination and `updateExamination` methods:

```
patientExamination.setPex_auscultation(dto.getPex_auscultation().name());
patientExamination.setPex_bowel_desc(dto.getPex_bowel_desc().name());
patientExamination.setPex_diuresis_desc(dto.getPex_diuresis_desc().name());
```

src/main/java/org/isf/examination/rest/ExaminationController.java
```diff
@@ -268,4 +199,50 @@ public ResponseEntity<List<PatientExaminationDTO>> getByPatientId(@PathVariable
...
+        if (newPatientExamination.getPex_temp() != null) {
+        	if (newPatientExamination.getPex_temp() < ExaminationParameters.TEMP_MIN || newPatientExamination.getPex_temp() > ExaminationParameters.TEMP_MAX) {
+             	throw new OHAPIException(new OHExceptionMessage("The temperature should be between "+ExaminationParameters.TEMP_MIN +" and "+ExaminationParameters.TEMP_MAX));
+            }
+        }
+	    if (newPatientExamination.getPex_sat() != null) {
+	    	if (newPatientExamination.getPex_sat() < ExaminationParameters.SAT_MIN || newPatientExamination.getPex_sat() > ExaminationParameters.SAT_MAX) {
+	         	throw new OHAPIException(new OHExceptionMessage("The saturation should be between "+ExaminationParameters.SAT_MIN+" and "+ExaminationParameters.SAT_MAX));
+	        } 
+	    }
```
mwithi, line 231: please fix indentation

mwithi (approved):

src/main/java/org/isf/examination/rest/ExaminationController.java
```diff
@@ -268,4 +199,50 @@ public ResponseEntity<List<PatientExaminationDTO>> getByPatientId(@PathVariable
...
+        	if (newPatientExamination.getPex_rr() < ExaminationParameters.RR_MIN || newPatientExamination.getPex_rr() > ExaminationParameters.RR_MAX) {
+            	throw new OHAPIException(new OHExceptionMessage("Respiratory rate should be between "+ExaminationParameters.RR_MIN+" and "+ExaminationParameters.RR_MAX));
+            }
+        }
+        if (newPatientExamination.getPex_diuresis() != null) {
+        	if (newPatientExamination.getPex_diuresis() < ExaminationParameters.DIURESIS_MIN || newPatientExamination.getPex_diuresis() > ExaminationParameters.DIURESIS_MAX) {
+            	throw new OHAPIException(new OHExceptionMessage("Diuresis should be between "+ExaminationParameters.DIURESIS_MIN+" and "+ExaminationParameters.DIURESIS_MAX));
+            }
+        }
+    }
```
mwithi, line 247: Now what is missing is that if **Diuresis**, **Bowel** and **Auscultation** are provided, they should be contained in their respective enumerations

mwithi (changes requested):
One indentation and checks on Diuresis, Bowel and Auscultation parameters needed

mwithi (approved):

src/main/java/org/isf/examination/rest/ExaminationController.java
mwithi: Found out that because of the casting in DTO, values different from enum are not accepted by default, so these check are not even needed, do you think we can keep them?

```
[27/oct./2023 16:19:54] [:] WARN - Resolved [org.springframework.http.converter.HttpMessageNotReadableException: JSON parse error: Cannot deserialize value of type `org.isf.examination.dto.Ausculation` from String "ciao": not one of the values accepted for Enum class: [bronchial, normal, rhonchi, crackles, wheezes, stridor]; nested exception is com.fasterxml.jackson.databind.exc.InvalidFormatException: Cannot deserialize value of type `org.isf.examination.dto.Ausculation` from String "ciao": not one of the values accepted for Enum class: [bronchial, normal, rhonchi, crackles, wheezes, stridor]<EOL> at [Source: (org.springframework.util.StreamUtils$NonClosingInputStream); line: 7, column: 23] (through reference chain: org.isf.examination.dto.PatientExaminationDTO["pex_auscultation"])]
```

Nevertheless, the enumerations **Diuresis**, **Bowel** and **Auscultation** should be shared within core, which means to move them into core and let API (and GUI) use them. Do you think we can do this now or in another PR?

ArnaudFonzam: yes I can remove it on api an add it on the core on this PR

mwithi: ok, let's try

mwithi (approved):

PR #372, opened 2023-10-24
OH2-243 | Fix user setting controller

ArnaudFonzam:
See OH2-243.


src/main/java/org/isf/menu/rest/UserSettingController.java
```diff
@@ -75,7 +80,16 @@ public UserSettingController(UserSettingMapper userSettingMapper, UserSettingMan
 	public ResponseEntity<UserSettingDTO> newUserSetting(@Valid @RequestBody UserSettingDTO userSettingDTO)
 					throws OHServiceException {
 		LOGGER.info("Attempting to create or update a UserSetting");
-		String userName = SecurityContextHolder.getContext().getAuthentication().getName();
+		String userName;
+		if (userSettingDTO.getUser() == null || userSettingDTO.getUser().equals("")) {
+			userName = SecurityContextHolder.getContext().getAuthentication().getName();
+		} else {
+			userName = userSettingDTO.getUser();
```
dbmalkovsky, line 87: why not just get the `userName` once before the first `if` instead and user that variable instead of going and getting it maybe 3 times?

ArnaudFonzam: Yes, you're right.

src/main/java/org/isf/menu/rest/UserSettingController.java
```diff
@@ -75,7 +80,15 @@ public UserSettingController(UserSettingMapper userSettingMapper, UserSettingMan
...
-		String userName = SecurityContextHolder.getContext().getAuthentication().getName();
+		String userName = userSettingDTO.getUser();
+		if (userName == null || userName.equals("")) {
+			userName = SecurityContextHolder.getContext().getAuthentication().getName();
+		} else {
+			if (userManager.getUserByName(userName) == null) {
+				throw new OHAPIException(new OHExceptionMessage("User not found."));
+			}
+		}
+		LOGGER.info("userName {}",userName);
```
mwithi, line 91: I don't get this change: BEFORE the setting was created always for the current user despite the passed `userName`; NOW if the `userName` is passed, it only checks if the users exists, which is good for admin, otherwise it just goes ahead with the creation/update, which is not good for other users, that can update any other user setting.

The issue states:

Aim of this issue is to fix the UserSettingController so it:

- allows to create and update the **current user’s own settings** `<-- not OK`
- allows admin create and update any user' settings `<-- OK`

So I think it is missing the check if the passed `userName` equals the current user.

I don't know how we could achieve **current user's own settings** while keeping **admin create and update any user' settings**, but I think we should enforce the first condition and maybe leave the second one as for now...

Make sense something like this?

```suggestion
		} else {
			if (!userName.equals(SecurityContextHolder.getContext().getAuthentication().getName()) {
			        throw new OHAPIException(new OHExceptionMessage("Not allowed."));
			}
		}
```

src/main/java/org/isf/menu/rest/UserSettingController.java
```diff
@@ -75,7 +80,15 @@ public UserSettingController(UserSettingMapper userSettingMapper, UserSettingMan
...
 		LOGGER.info("Attempting to create or update a UserSetting");
-		String userName = SecurityContextHolder.getContext().getAuthentication().getName();
+		String userName = userSettingDTO.getUser();
+		if (userName == null || userName.equals("")) {
+			userName = SecurityContextHolder.getContext().getAuthentication().getName();
+		} else {
+			if (userManager.getUserByName(userName) == null) {
+				throw new OHAPIException(new OHExceptionMessage("User not found."));
+			}
+		}
```
mwithi, line 90: I don't get this change: BEFORE the setting was created always for the current user despite the passed `userName`; NOW if the `userName` is passed, it only checks if the users exists, which is good for admin, otherwise it just goes ahead with the creation/update, which is not good for other users, that can update any other user setting.

The issue states:

Aim of this issue is to fix the UserSettingController so it:

- allows to create and update the **current user’s own settings** `<-- not OK`
- allows admin create and update any user' settings `<-- OK`

So I think it is missing the check if the passed `userName` equals the current user.

I don't know how we could achieve **current user's own settings** while keeping **admin create and update any user' settings**, but I think we should enforce the first condition and maybe leave the second one as for now...

Make sense something like this?

```suggestion
		} else {
			if (!userName.equals(SecurityContextHolder.getContext().getAuthentication().getName()) {
			        throw new OHAPIException(new OHExceptionMessage("Not allowed."));
			}
		}
```

src/main/java/org/isf/menu/rest/UserSettingController.java
mwithi, line 91:
```suggestion
		LOGGER.info("userName {}", userName);
```

src/main/java/org/isf/menu/rest/UserSettingController.java
ArnaudFonzam: I think with the script that you provide it is not possible for the admin to create or update the setting of another user because the name in userSettingDTO will be different to the current use who is admin Or we can set user Not null in userSettingDTO then if the current user is different to user pass in userSettingDTO we can check if that current user is admin if yes we allow them to make action(create or update) if not we Not allow them. if the current is the user pass is userSettingDTO we allow also. WDTY?

mwithi: I think yes, please make the change and we see. Thanks!

src/main/java/org/isf/menu/rest/UserSettingController.java
```diff
@@ -81,12 +81,11 @@ public ResponseEntity<UserSettingDTO> newUserSetting(@Valid @RequestBody UserSet
...
-			userName = SecurityContextHolder.getContext().getAuthentication().getName();
-		} else {
-			if (!userName.equals(SecurityContextHolder.getContext().getAuthentication().getName())) {
-			        throw new OHAPIException(new OHExceptionMessage("Not allowed."));
-			}
+		final String ADMIN = "admin";
+		if (!userName.equals(SecurityContextHolder.getContext().getAuthentication().getName())) {
+		    if (!userName.equals(ADMIN)) {
+		    	throw new OHAPIException(new OHExceptionMessage("Not allowed."));
+		    }
```
mwithi, line 88: Could you combine these two conditions?

mwithi (approved):
LGTM. Thanks a lot!

PR #401, opened 2023-11-22
OH2-246 Fix

mwithi:
See OH2-246.

It seems that to update a contributor PR with an automatically updated openapi/oh.yaml specs has too many donwsides in terms of workflow and permissions.

So I've replaced `gr2m/create-or-update-pull-request-action@v1` with https://github.com/oasdiff/oasdiff-action, a tool that just checks the `diff `(or `breaking-changes` in this case) between two provided spec: the first one is the one on the repo, the second one (revision) is the one generated "on the fly" with the Maven `springdoc-openapi plugin`.

If breaking changes are detected, the GitHub action will pass, otherwise the contributor is requested to generate locally (command `mvn springdoc-openapi:generate`) and push the new openapi spec file `oh.yaml`.

In this way the `openapi/oh.yaml` will be always aligned with the code.

**Note**: Maven `springdoc-openapi:generate` target needs openhospital-api up and running in order to work.


openapi/oh.yaml
```diff
@@ -8117,30 +8085,31 @@ components:
...
+                          type: boolean
+                        location:
                           type: object
                           properties:
-                            width:
+                            x:
                               type: number
                               format: double
-                            height:
+                            "y":
```
dbmalkovsky, line 8096: Why is `y` in quotes and `x` isn't?   It seems consistent that way through the file.

mwithi: I noticed that but really I don't know... this yamel is generated by the `springdoc-openapi` maven plugin.
but if you launch API and open `http://localhost:8080/v3/api-docs.yaml` it is same way 🤷‍♂️

dbmalkovsky (approved):
LGTM given my limited understanding 😃

PR #411, opened 2023-12-13
OH2-255 Revise and align permissions for CORE, API and UI

mwithi:
See OH2-255.

- [x] remove duplicates
- [x] add missing
- [x] refactor some controllers and endpoints (resources in plurals)
- [x] made all permissions spelled like endpoint
- [x] sort in alphabetical order, then CRUD order
- [x] remove PATCH methods (we are not using as for now, maybe in future)
- [x] move `/discharges` to `/admissions/discharges` endpoint and remove related permissions (`admissions.*` ones will be used)
- [x] update oh.yaml

PR #415, opened 2023-12-19
Update PR #411

npasquetto:
Changes for #411 

Modify duplicate @PostMapping in VisitsController
Update oh.yaml
Fix Tests

### DIFF
--- a/src/main/java/org/isf/visits/rest/VisitsController.java
+++ b/src/main/java/org/isf/visits/rest/VisitsController.java
@@ -96,7 +96,7 @@ public class VisitsController {
 	 * @return an error if there are some problem, the visitor id (Integer) otherwise
 	 * @throws OHServiceException
 	 */
-	@PostMapping(value = "/visit", produces = MediaType.APPLICATION_JSON_VALUE)
+	@PostMapping(value = "/visits", produces = MediaType.APPLICATION_JSON_VALUE)
 	public ResponseEntity<VisitDTO> newVisit(@RequestBody VisitDTO newVisit) throws OHServiceException {
 		LOGGER.info("Create Visit: {}", newVisit);
 		Visit visitD = mapper.map2Model(newVisit);
@@ -111,7 +111,7 @@ public class VisitsController {
 	 * @return an error message if there are some problem, ok otherwise
 	 * @throws OHServiceException
 	 */
-	@PostMapping(value = "/visits", produces = MediaType.APPLICATION_JSON_VALUE)
+	@PostMapping(value = "/visits/insertList", produces = MediaType.APPLICATION_JSON_VALUE)
 	public ResponseEntity<Boolean> newVisits(@RequestBody List<VisitDTO> newVisits) throws OHServiceException {
 		LOGGER.info("Create Visits");
 		List<Visit> listVisits = mapper.map2ModelList(newVisits);
--- a/src/main/java/org/isf/visits/rest/VisitsController.java
+++ b/src/main/java/org/isf/visits/rest/VisitsController.java
@@ -96,7 +96,7 @@ public class VisitsController {
 	 * @return an error if there are some problem, the visitor id (Integer) otherwise
 	 * @throws OHServiceException
 	 */
-	@PostMapping(value = "/visits", produces = MediaType.APPLICATION_JSON_VALUE)
+	@PostMapping(value = "/visit", produces = MediaType.APPLICATION_JSON_VALUE)
 	public ResponseEntity<VisitDTO> newVisit(@RequestBody VisitDTO newVisit) throws OHServiceException {
 		LOGGER.info("Create Visit: {}", newVisit);
 		Visit visitD = mapper.map2Model(newVisit);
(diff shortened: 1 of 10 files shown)


src/test/java/org/isf/lab/data/LaboratoryHelper.java
```diff
@@ -53,7 +57,25 @@ public static Laboratory setup(Integer id) throws OHException {
...
+				.registerModule(new JavaTimeModule())
+				.writeValueAsString(body);
+		} catch (JsonProcessingException e) {
+			e.printStackTrace();
+		}
+		return null;
+	}
+
+	public static String asJsonString(LabWithRowsDTO body) {
+		// TODO Auto-generated method stub
```
dbmalkovsky, line 72: Please remove this comment

src/main/java/org/isf/visits/rest/VisitsController.java
```diff
@@ -96,7 +96,7 @@ public ResponseEntity<List<VisitDTO>> getVisit(@PathVariable("patID") int patID)
 	 * @return an error if there are some problem, the visitor id (Integer) otherwise
 	 * @throws OHServiceException
 	 */
-	@PostMapping(value = "/visits", produces = MediaType.APPLICATION_JSON_VALUE)
+	@PostMapping(value = "/visit", produces = MediaType.APPLICATION_JSON_VALUE)
```
dbmalkovsky, line 99: Didn't a recent update make this plural?  why change it back?

src/test/java/org/isf/vactype/rest/VaccineTypeControllerTest.java
```diff
@@ -88,7 +88,7 @@ void closeService() throws Exception {
 
 	@Test
 	public void testGetVaccineType_200() throws Exception {
-		String request = "/vaccinetype";
+		String request = "/vaccinetypes";
```
dbmalkovsky, line 91: again; didn't a recent change make these plural?

src/main/java/org/isf/visits/rest/VisitsController.java
npasquetto: For two reasons:
1) It is not possible to have two @PostMapping with the same value
2) In this method we want to create a single visit, on the other one (the one mapped on "/visits") we want to create a list of visit, so for coherence this method shoud be mapped on "/visit"

src/test/java/org/isf/vactype/rest/VaccineTypeControllerTest.java
npasquetto: Yes, but these tests were left in the singular, so I updated them.

src/main/java/org/isf/visits/rest/VisitsController.java
mwithi: I see, I didn't realize that! Nevertheless, I think this could be error prone... could we change it to something totally different, like `POST /visits/list`?

(FWIW and IMHO, I prefer singular endpoints for a number of reasons (plurals may differ (visit-s, laborator-ies, ecc...) and make everything more verbose, but I decided to follow the standard we have so far in the repo)

src/test/java/org/isf/vactype/rest/VaccineTypeControllerTest.java
mwithi: Yes @dbmalkovsky this change is a fix on a missing change in the previous PR, correct @npasquetto ?

src/test/java/org/isf/lab/data/LaboratoryHelper.java
```diff
@@ -35,41 +36,64 @@
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
+import com.fasterxml.jackson.module.paramnames.ParameterNamesModule;
```
mwithi, line 41: Are these modules needed only for Lab tests or potentially for every test?

src/test/java/org/isf/vactype/rest/VaccineTypeControllerTest.java
npasquetto: right

src/main/java/org/isf/visits/rest/VisitsController.java
npasquetto: I've reused the `insertList` pattern, already used in oh-api for Laboratories;
please, check last commit.

src/test/java/org/isf/lab/data/LaboratoryHelper.java
npasquetto: I prefer to load them only if needed

src/test/java/org/isf/vactype/rest/VaccineTypeControllerTest.java
mwithi: it's ok for me then

src/main/java/org/isf/visits/rest/VisitsController.java
mwithi: yes, it's ok for me then

mwithi (approved):
LGTM. Thanks a lot!

npasquetto:
> LGTM. Thanks a lot!

Good! so if we can approve this PR, we then be able to evaluate the #411

mwithi:
> > LGTM. Thanks a lot!
> 
> Good! so if we can approve this PR, we then be able to evaluate the #411

Done!

PR #416, opened 2023-12-20
OP-312 Telemetry Module

mwithi:
See OP-312.


rsc/telemetry.properties
```diff
@@ -0,0 +1,11 @@
+# number of second to check if it should send a message
```
dbmalkovsky, line 1:
```suggestion
# number of seconds to check if it should send a message

```

rsc/telemetry.properties
```diff
@@ -0,0 +1,11 @@
+# number of second to check if it should send a message
+telemetry.daemon.thread.loop.seconds=14400
```
dbmalkovsky, line 2: Do you mean to mix line endings?

dbmalkovsky (approved):
LGTM with a few comments

rsc/telemetry.properties
mwithi: strange! let me fix that

PR #422, opened 2024-01-07
OP-1165-fix-circular-reference-in-jpa-config

mwithi:
Just for testing OP-1165.

Possibly, merging https://github.com/informatici/openhospital-api/pull/422/commits/d29a394ab0610aa3a23e56ff45573f6246af124f

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -15,7 +15,7 @@ jobs:
           persist-credentials: false
 
       - name: Checkout core
-        run: git clone --depth=50 --branch=develop https://github.com/informatici/openhospital-core.git openhospital-core
+        run: git clone --depth=50 --branch=OP-1165-fix-circular-reference-in-jpa-config https://github.com/informatici/openhospital-core.git openhospital-core
 
       - name: Install core
         run: cd openhospital-core && mvn install -DskipTests=true && cd ..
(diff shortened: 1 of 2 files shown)


.github/workflows/maven.yml
```diff
@@ -15,7 +15,7 @@ jobs:
           persist-credentials: false
 
       - name: Checkout core
-        run: git clone --depth=50 --branch=develop https://github.com/informatici/openhospital-core.git openhospital-core
+        run: git clone --depth=50 --branch=OP-1165-fix-circular-reference-in-jpa-config https://github.com/informatici/openhospital-core.git openhospital-core
```
dbmalkovsky, line 18: doesn't this need to be changed before merge?

dbmalkovsky (approved):
LGTM after comment is addressed

.github/workflows/maven.yml
mwithi: yes, of course, if the change to `src/main/java/org/isf/config/SecurityConfig.java` is also neeed, otherwise we can just close this PR becuase the problem was actually in core only.

mwithi:
Closing because not needed.

> Just for testing [OP-1165](https://openhospital.atlassian.net/browse/OP-1165).
> 
> Possibly, merging [d29a394](https://github.com/informatici/openhospital-api/commit/d29a394ab0610aa3a23e56ff45573f6246af124f)

Closing because tests are ok and change in d29a394](https://github.com/informatici/openhospital-api/commit/d29a394ab0610aa3a23e56ff45573f6246af124f is not really needed.

PR #429, opened 2024-01-31
OP-336 Convert test to junit5 and test contextLoads() method

mwithi:
See OP-336.

Not sure if to have contextLoads() test though, by the moment it is slows down the whole test suite

### DIFF
--- a/src/test/java/org/isf/OpenHospitalApiApplicationTests.java
+++ b/src/test/java/org/isf/OpenHospitalApiApplicationTests.java
@@ -21,6 +21,7 @@
  */
 package org.isf;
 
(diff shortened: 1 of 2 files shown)


dbmalkovsky (approved):
LGTM

emecas (commented):
That is exactly what I did, to verified it was working on,  removing the comments from the test, no chance to test against the original issue running it on Eclipse.

I appreciate your effort in verifying that the loading context is working correctly. However, since this test does not provide any additional value and may actually slow down the tests execution, it's recommended to remove it. The loading context is implicitly tested by other tests that run as SpringBootTest.

PR #438, opened 2024-04-06
OP-1207 move to Spring Boot 3.2.x

dbmalkovsky:
See OP-1207

Notes:

See `src/main/java/org/isf/config/SecurityConfig.java` there are still many deprecated methods in this file that will be removed in a future version of Spring.  I only made the necessary changes for methods that have already been removed.

There are four tests that failed and changes were made in the expected return types to get them to work.  Search for `// TODO .` to find the changes. The TODO line is what was there and the following line is the change I made to make it work.


pom.xml
```diff
@@ -15,9 +15,8 @@
     <description>Open Hospital Spring Boot REST API</description>
 
     <properties>
-        <oh-core.version>1.14.0-SNAPSHOT</oh-core.version>
+        <oh-core.version>1.14.1.SNAPSHOT</oh-core.version>
```
mwithi, line 18:
```suggestion
        <oh-core.version>1.14.1-SNAPSHOT</oh-core.version>
```

mwithi (changes requested):
It looks like swagger UI is no longer working as expected in the README.
Please try to fix it...
(Note: GUI and UI seem working fine instead)

dbmalkovsky:
In the README `org.springframework.boot.loader.JarLauncher` should be `org.springframework.boot.loader.launch.JarLauncher`.  The class has moved.  I also updated the pom.xml file with two jars that match Spring Boot 3.2.x.   I can start Spring Boot but the process dies because of the JWT Token being missing.   Right now I don't have the time and/or patience to figure out the security stuff; sorry.

mwithi:
> In the README `org.springframework.boot.loader.JarLauncher` should be `org.springframework.boot.loader.launch.JarLauncher`. The class has moved. I also updated the pom.xml file with two jars that match Spring Boot 3.2.x. I can start Spring Boot but the process dies because of the JWT Token being missing. Right now I don't have the time and/or patience to figure out the security stuff; sorry.

You was almost there: the JWT token can be any SHA-256 string of your choice. With that and your last commit, it works like a charme! Thanks! 👏

mwithi (approved):

PR #439, opened 2024-04-22
Chore: update to SpringBoot 3.2.5

dbmalkovsky:
This change matches changes in CORE.

Not sure about the intermediate commits; I tried resetting things but nothing seems to help.

PR #440, opened 2024-04-22
Fix OpenAPI

mwithi:
- Remove duplicated Controller
- Fix launcher
- Add Hibernate dialect (disable auto-detecting, despite discouraged from Hibernate 6+)
- Remove `fail-on-diff: true` because default is `true`


mwithi:
superseded by https://github.com/informatici/openhospital-api/pull/439

PR #441, opened 2024-05-06
OP-1234: fix tests related to CORE changes in OP-1210

dbmalkovsky:
See OP-1234

I don't understand the other commits in this pull request.  I have reset the head of develop; re-fetched the repo twice, and tried other things.   I just give up.

Maybe if this gets merged things will straighten themselves out and I can look at the other SpringBoot issue.


mwithi (approved):
LGTM. Thanks a lot!

PR #449, opened 2024-06-06
OH2-310 Add missing supDeleted field in SupplierDto

mwithi:
See OH2-310.

### DIFF
--- a/src/main/java/org/isf/supplier/dto/SupplierDTO.java
+++ b/src/main/java/org/isf/supplier/dto/SupplierDTO.java
@@ -21,9 +21,8 @@
  */
 package org.isf.supplier.dto;
 
-import jakarta.validation.constraints.NotNull;
-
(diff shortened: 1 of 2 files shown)


src/main/java/org/isf/supplier/dto/SupplierDTO.java
```diff
@@ -21,9 +21,8 @@
  */
 package org.isf.supplier.dto;
 
-import jakarta.validation.constraints.NotNull;
-
 import io.swagger.v3.oas.annotations.media.Schema;
+import jakarta.validation.constraints.NotNull;
```
dbmalkovsky: I think the original import order was correct.

dbmalkovsky (commented):
Made a single comment about import order.

dbmalkovsky (approved):
LGTM

PR #462, opened 2024-08-07
OH2-363 | Remove APIs for CUD on Permissions

SilverD3:
See OH2-363

### DIFF
--- a/src/main/java/org/isf/permissions/dto/PermissionDTO.java
+++ b/src/main/java/org/isf/permissions/dto/PermissionDTO.java
@@ -21,8 +21,8 @@
  */
 package org.isf.permissions.dto;
 
-import io.swagger.v3.oas.annotations.media.Schema;
 import jakarta.validation.constraints.NotNull;
+import io.swagger.v3.oas.annotations.media.Schema;
 
 public class PermissionDTO extends LitePermissionDTO {
 
--- a/src/main/java/org/isf/permissions/dto/PermissionDTO.java
+++ b/src/main/java/org/isf/permissions/dto/PermissionDTO.java
@@ -21,11 +21,8 @@
  */
 package org.isf.permissions.dto;
 
-import java.util.List;
-
-import jakarta.validation.constraints.NotNull;
-
 import io.swagger.v3.oas.annotations.media.Schema;
+import jakarta.validation.constraints.NotNull;
 
 public class PermissionDTO extends LitePermissionDTO {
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/isf/permissions/dto/PermissionDTO.java
```diff
@@ -21,11 +21,8 @@
  */
 package org.isf.permissions.dto;
 
-import java.util.List;
-
-import jakarta.validation.constraints.NotNull;
-
 import io.swagger.v3.oas.annotations.media.Schema;
+import jakarta.validation.constraints.NotNull;
```
dbmalkovsky, line 24: Please follow the project standards for ordering of imports; see https://github.com/informatici/openhospital-core/blob/develop/.ide-settings/eclipse/OpenHospital.importorder

src/main/java/org/isf/permissions/rest/PermissionController.java
```diff
@@ -22,35 +22,23 @@
 package org.isf.permissions.rest;
 
 import java.util.List;
-
-import org.isf.menu.manager.UserGroupManager;
-import org.isf.menu.model.UserGroup;
+import io.swagger.v3.oas.annotations.security.SecurityRequirement;
```
dbmalkovsky: These imports should be located where they were in the original file.

src/main/java/org/isf/permissions/mapper/PermissionMapper.java
```diff
@@ -39,27 +37,11 @@ public PermissionMapper() {
...
-		dto.setUserGroupIds(userGroupCodes);
-		return dto;
-	}
-
-	private List<String> extractUserGroupCodes(Permission fromObj) {
-		if (fromObj.getGroupPermission() == null) {
-			return new ArrayList<>();
-		}
-		return fromObj.getGroupPermission().stream().map(gp -> gp.getUserGroup().getCode()).collect(Collectors.toList());
+		return modelMapper.map(fromObj, PermissionDTO.class);
```
dbmalkovsky, line 45: same comment as above.

It would be a great service to everyone if there were some tests for the `permissions` package in this repo;  Currently there are none.

SilverD3: My next PR will be about permissions endpoints update. It will bring permissions tests

mwithi: @SilverD3 why is not possible to have permissions JUnit tests in this PR itself?

SilverD3: It's possible. It's just that there's another issue (OP-1309) I've started in which I'm going to modify the permissions endpoints, and I think it would be more practical to write the tests in that other task.

mwithi: I see... ok then!

mwithi (approved):
LGTM. Thanks a lot!

PR #466, opened 2024-09-09
OH2-378 | Fix 400 Error When creating Supplier

ArnaudFonzam:
see [OH2-378](https://github.com/informatici/openhospital-core/pull/1403)

[OH2-378]: https://openhospital.atlassian.net/browse/OH2-378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/isf/supplier/dto/SupplierDTO.java
+++ b/src/main/java/org/isf/supplier/dto/SupplierDTO.java
@@ -1,6 +1,6 @@
 /*
  * Open Hospital (www.open-hospital.org)
- * Copyright © 2006-2023 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
+ * Copyright © 2006-2024 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
  *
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/supplier/dto/SupplierDTO.java
mwithi, line 1: Why you don't need this information in the DTO?

ArnaudFonzam: Because I think this property is manage on back-end Side and the deletion action is not available on front-end and back-end side.

mwithi: The soft-deletion should anyway start by the frontend, but it is true that we still miss the API for that, wouldn't you mind to create an issue for that? (low priority) Thanks!

mwithi (approved):
LGTM. Thanks a lot!

PR #471, opened 2024-09-25
OP-1337 | Add token refresh mechanism

mwithi:
See OP-1337.

- Added also test `testAuthenticateUser_Success`
- Added `TODO` comment for `testAuthenticateUser_Failure` (I tried for hours without success, needs more investigation)


src/main/java/org/isf/login/rest/LoginController.java
```diff
@@ -97,6 +115,26 @@ public ResponseEntity<LoginResponse> authenticateUser(@Valid @RequestBody LoginR
...
+	}
+
+	@PostMapping("/auth/refresh-token")
+	public ResponseEntity< ? > refreshToken(@RequestBody TokenRefreshRequest request) {
+		String refreshToken = request.getRefreshToken();
+
+		try {
+			if (tokenProvider.validateToken(refreshToken) == TokenValidationResult.VALID) {
+				String username = tokenProvider.getUsernameFromToken(refreshToken);
+				Authentication authentication = tokenProvider.getAuthenticationByUsername(username); // Retrieve user details
```
mwithi, line 128:
```suggestion
				Authentication authentication = tokenProvider.getAuthenticationByUsername(username);
```

src/main/java/org/isf/login/rest/LoginController.java
```diff
@@ -49,10 +52,11 @@
 import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RestController;
 
+import io.jsonwebtoken.JwtException;
 import io.swagger.v3.oas.annotations.security.SecurityRequirement;
 import io.swagger.v3.oas.annotations.tags.Tag;
 
-@RestController(value = "/auth")
```
mwithi, line 55: @SilverD3 the `value` parameter for `@RestController` annotation is not really needed, should we remove it in all controllers?

SilverD3: Yes, it has no effect. I suppose when it was added, it was supposed to behave like `@RequestMapping(value = "/auth")`.

src/main/java/org/isf/security/jwt/TokenProvider.java
```diff
@@ -71,15 +73,21 @@ public class TokenProvider implements Serializable {
...
 	@PostConstruct
 	public void init() {
 		String secret = env.getProperty("jwt.token.secret");
 		LOGGER.info("Initializing JWT key with secret: {}", secret);
 		byte[] keyBytes = secret.getBytes(StandardCharsets.UTF_8);
 		this.key = Keys.hmacShaKeyFor(keyBytes);
 
-		this.tokenValidityInMilliseconds = 1000L * 6000;
-		this.tokenValidityInMillisecondsForRememberMe = 1000L * 6000;
+		// 15 minutes (900,000 milliseconds)
```
SilverD3: 15 minutes is too short I think. Maybe 30 minutes or 1 hour.

src/main/java/org/isf/security/jwt/TokenProvider.java
```diff
@@ -71,15 +73,21 @@ public class TokenProvider implements Serializable {
...
+		this.tokenValidityInMilliseconds = 1000L * 60 * 15;
+
+		// 7 days (604,800,000 milliseconds)
+		this.tokenValidityInMillisecondsForRememberMe = 1000L * 60 * 60 * 24 * 7;
```
SilverD3: Toooo long! 3 days or less is enough I think.

SilverD3 (changes requested):
### General consideration
1. I think the refresh token should be refreshed also when the token is refresh, and of course the old refresh token should be revoked.
2. Is there a mechanism to revoke the refresh token when the user logs out? (I suppose even the token is not revoked when user logs out, right?)

src/main/java/org/isf/login/rest/LoginController.java
```diff
@@ -74,12 +78,26 @@ public class LoginController {
 
 	private static final Logger LOGGER = LoggerFactory.getLogger(LoginController.class);
 
+	@Autowired
```
dbmalkovsky: What is the purpose of this?   `@Autowired` should be removed.    SpringBoot knows what to do with this constructor.

dbmalkovsky: And all the other `@Autowired` in line 64-76 can be removed too.

src/test/java/org/isf/security/TokenProviderTest.java
```diff
@@ -22,25 +22,37 @@
 package org.isf.security;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.assertThrows;
```
dbmalkovsky:
```suggestion
import static org.junit.jupiter.api.Assertions.assertThrows;
```

dbmalkovsky: Why use junit 4 code when everything else is junit 5+ ?

dbmalkovsky (commented):
Left some comments.   It was also be good to add javadoc to some of the new API calls.

src/main/java/org/isf/login/rest/LoginController.java
mwithi: yes, and it could even bring problem as far as I'm understanding, don't you mind to open an issue to remove all of them?

src/main/java/org/isf/security/jwt/TokenProvider.java
mwithi: Thanks! I will make them configurable anyway

src/main/java/org/isf/security/jwt/TokenProvider.java
mwithi: Thanks! I will make them configurable anyway

src/main/java/org/isf/login/rest/LoginController.java
mwithi: you're right, just left after some tries... for this reason maybe I had problem in testing the login failure... I will try again.
Thanks!

src/test/java/org/isf/security/TokenProviderTest.java
mwithi: automatic import... fixed, thanks!

mwithi:
> ### General consideration
> 1. I think the refresh token should be refreshed also when the token is refresh, and of course the old refresh token should be revoked.
> 2. Is there a mechanism to revoke the refresh token when the user logs out? (I suppose even the token is not revoked when user logs out, right?)

1. it's very possible to do that
2. a token can be only blacklisted, not revoked (it works until is expired or the server changes its key) but blacklist mechanism is not yet implemented so far

Should I re-generate also the refresh token? If so, it would increase the number of tokens to be blacklisted in the future

mwithi:
> Left some comments. It was also be good to add javadoc to some of the new API calls.

The only new API is `/auth/refresh-token`, but I think it's self-explanatory, or not?

src/main/java/org/isf/login/rest/LoginController.java
SilverD3: Yes, you can definitely remove them, I mean if it's okay for @dbmalkovsky too.

mwithi: Not in this PR anyway

SilverD3:
> Should I re-generate also the refresh token? If so, it would increase the number of tokens to be blacklisted in the future

Yes. We will be keeping those tokens blacklisted until they expire and then we could delete them to free memory.

SilverD3:
> a token can be only blacklisted, not revoked (it works until is expired or the server changes its key) but blacklist mechanism is not yet implemented so far

Are you going to implement it in this PR or in another one?

mwithi:
> > a token can be only blacklisted, not revoked (it works until is expired or the server changes its key) but blacklist mechanism is not yet implemented so far
> 
> Are you going to implement it in this PR or in another one?

Another one

mwithi:
> > Should I re-generate also the refresh token? If so, it would increase the number of tokens to be blacklisted in the future
> 
> Yes. We will be keeping those tokens blacklisted until they expire and then we could delete them to free memory.

Done

src/main/java/org/isf/login/rest/LoginController.java
```diff
@@ -97,6 +108,27 @@ public ResponseEntity<LoginResponse> authenticateUser(@Valid @RequestBody LoginR
 			LOGGER.error("Unable to log user login in the session_audit table");
 		}
 
-		return ResponseEntity.ok(new LoginResponse(jwt, userDetails));
+		return ResponseEntity.ok(new LoginResponse(jwt, refreshToken, userDetails));
 	}
+
+	@PostMapping("/auth/refresh-token")
+	public ResponseEntity< ? > refreshToken(@RequestBody TokenRefreshRequest request) {
```
SilverD3: Returning an unknown response type could be hard to implement in the FE and could lead to error. I think you can return `LoginResponse` and if an error occurs, just throw an exception.

src/main/java/org/isf/security/jwt/TokenProvider.java
```diff
@@ -71,15 +73,21 @@ public class TokenProvider implements Serializable {
...
 		byte[] keyBytes = secret.getBytes(StandardCharsets.UTF_8);
 		this.key = Keys.hmacShaKeyFor(keyBytes);
 
-		this.tokenValidityInMilliseconds = 1000L * 6000;
-		this.tokenValidityInMillisecondsForRememberMe = 1000L * 6000;
+		// 30 minutes (900,000 milliseconds)
+		this.tokenValidityInMilliseconds = 1000L * 60 * 30;
+
+		// 3 days (604,800,000 milliseconds)
+		this.tokenValidityInMillisecondsForRememberMe = 1000L * 60 * 60 * 24 * 3;
```
SilverD3, line 90: You said that you're going to make them configurable

SteveGT96:
> > ### General consideration
> > 
> > 1. I think the refresh token should be refreshed also when the token is refresh, and of course the old refresh token should be revoked.
> > 2. Is there a mechanism to revoke the refresh token when the user logs out? (I suppose even the token is not revoked when user logs out, right?)
> 
> 1. it's very possible to do that
> 2. a token can be only blacklisted, not revoked (it works until is expired or the server changes its key) but blacklist mechanism is not yet implemented so far
> 
> Should I re-generate also the refresh token? If so, it would increase the number of tokens to be blacklisted in the future

We could use the same access token to refresh token. The implementation could look like follows:

- When the user login we generate a token and save it.
- The endpoint to refresh the tocken accept a payload containing the uid and the token. We regenerate the token if the token match the current user access token and if the duration between the token expiration and the token issued date dosn't exceed the refresh expiration.
- When the user logs out we just remove the access token.

It could be more easy if we were using redis:
- When the user logging, we save the token in redis using for example the key "token:[uid]"(eg: "token:admin") and set the expiration to the refresh token expiration.
- On each request with the bearer auth, after validation the token, we check if it still exist in redis by comparing the token to the db element with key "token:[uid]". If they not match,probably the token has been revoked.
- When refreshing the user token using [uid, access_token] as payload, we just make sure that the element in db with the key "token:[uid]" is not null and is equal to the access_token. If OK then we regenerate a new token and save it again(in the key 'token:[uid]"), then return it.
- When the user logs out, then we just remove the token in the redis db(corresponding to the key 'token:[uid]'

What do you think ?

mwithi:
> What do you think ?

Hi @SteveGT96, not sure if here is the correct place where to discuss this:
1) the refresh token mechanism is quite standard and we don't need to reinvent it
2) at the moment we don't persist yet the tokens in a DB or Redis

I would move the discussion under the OP-1342.

src/main/java/org/isf/login/rest/LoginController.java
SilverD3: Please open an issue on Jira for that if not yet

SilverD3:
> What do you think ?

@SteveGT96 we could discuss about this when we'll be in the enhancement phase. Adding too much technologies make it difficult to maintain and deploy the app

SteveGT96:
> > What do you think ?
> 
> @SteveGT96 we could discuss about this when we'll be in the enhancement phase. Adding too much technologies make it difficult to maintain and deploy the app

Roger that.

src/main/java/org/isf/login/rest/LoginController.java
mwithi: Created OH2-403 for this.

SilverD3 (changes requested):

src/main/java/org/isf/login/rest/LoginController.java
mwithi: you mean this way?

image

src/main/java/org/isf/security/jwt/TokenProvider.java
mwithi: yes, in another issue

src/main/java/org/isf/login/rest/LoginController.java
SilverD3: Yeah

mwithi: alright, thansk!
what about the `e.printStackTrace();` at line 101?

SilverD3 (approved):

PR #484, opened 2024-10-22
worfklow: trying to build against a core branch with same name (develop otherwise)

mwithi:
In order to have checks to pass when an API PR depends on a CORE PR, let's try to checkout the corresponding branch on CORE, if any, otherwise default to develop (like before)


dbmalkovsky (approved):
LGTM

PR #485, opened 2024-10-25
OP-1312 | Create 'oh_settings' controller for CRUD for enabled users

SilverD3:
See OP-1312

This PR includes also integration tests for SettingController


src/test/java/org/isf/settings/data/SettingHelper.java
```diff
@@ -0,0 +1,95 @@
+package org.isf.settings.data;
```
dbmalkovsky, line 1: Missing license header.

PR #494, opened 2024-10-30
fix: OH2-413 - Fix testGetAgeTypeByIndex_200 | OH2-414 - Fix 3 UpdateExamTests

SilverD3:
See OH2-413 and OH2-414


mwithi (approved):
LGTM. Thanks a lot!

PR #496, opened 2024-10-30
OP-1043 | Implement soft deletion

SteveGT96:
See [OP-1043](https://openhospital.atlassian.net/browse/OP-1043)

[OP-1043]: https://openhospital.atlassian.net/browse/OP-1043?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/isf/usergroups/dto/UserGroupDTO.java
+++ b/src/main/java/org/isf/usergroups/dto/UserGroupDTO.java
@@ -31,36 +31,38 @@ import io.swagger.v3.oas.annotations.media.Schema;
 
 public class UserGroupDTO {
 
+	@Schema(description = "List of group's permissions")
+	List<PermissionDTO> permissions;
 	@NotNull
 	@Schema(description = "Name of the group (must be unique)", example = "labo", maxLength = 50)
 	private String code;
-
 	@Schema(description = "The description of the group", example = "Staff members working in the laboratory", maxLength = 128)
 	private String desc;
-
-	@Schema(description = "List of group's permissions")
-	List<PermissionDTO> permissions;
+	@Schema(description = "Whether the group has been soft deleted or not", example = "false")
+	private boolean deleted;
 
 	public UserGroupDTO() {
 	}
 
 	public UserGroupDTO(String code, String desc) {
+		this(code, desc, false);
+	}
+
+	public UserGroupDTO(String code, String desc, boolean deleted) {
 		this.code = code;
 		this.desc = desc;
+		this.deleted = deleted;
 	}
 
 	public String getCode() {
 		return this.code;
 	}
-
-	public String getDesc() {
-		return this.desc;
-	}
-
 	public void setCode(String code) {
 		this.code = code;
 	}
-
+	public String getDesc() {
+		return this.desc;
+	}
 	public void setDesc(String desc) {
 		this.desc = desc;
 	}
@@ -71,5 +73,12 @@ public class UserGroupDTO {
 
 	public void setPermissions(List<PermissionDTO> permissions) {
 		this.permissions = permissions;
+	}
+
+	public boolean isDeleted() {
+		return deleted;
+	}
+	public void setDeleted(boolean deleted) {
+		this.deleted = deleted;
 	}
 }
--- a/src/main/java/org/isf/usergroups/dto/UserGroupDTO.java
+++ b/src/main/java/org/isf/usergroups/dto/UserGroupDTO.java
@@ -31,8 +31,6 @@ import io.swagger.v3.oas.annotations.media.Schema;
 
 public class UserGroupDTO {
 
-	@Schema(description = "List of group's permissions")
-	List<PermissionDTO> permissions;
 	@NotNull
 	@Schema(description = "Name of the group (must be unique)", example = "labo", maxLength = 50)
 	private String code;
@@ -40,6 +38,9 @@ public class UserGroupDTO {
 	private String desc;
 	@Schema(description = "Whether the group has been soft deleted or not", example = "false")
 	private boolean deleted;
+
+	@Schema(description = "List of group's permissions")
+	private List<PermissionDTO> permissions;
 
 	public UserGroupDTO() {
 	}
@@ -78,6 +79,7 @@ public class UserGroupDTO {
 	public boolean isDeleted() {
 		return deleted;
 	}
+
 	public void setDeleted(boolean deleted) {
 		this.deleted = deleted;
 	}
--- a/src/main/java/org/isf/users/dto/UserDTO.java
+++ b/src/main/java/org/isf/users/dto/UserDTO.java
@@ -89,6 +89,7 @@ public class UserDTO {
 	public boolean isDeleted() {
 		return deleted;
 	}
+	
 	public void setDeleted(boolean deleted) {
 		this.deleted = deleted;
 	}
--- a/src/main/java/org/isf/users/dto/UserDTO.java
+++ b/src/main/java/org/isf/users/dto/UserDTO.java
@@ -43,15 +43,22 @@ public class UserDTO {
 
 	@Schema(description = "The user's description", example = "Lab chief technician", maxLength = 128)
 	private String desc;
+	@Schema(description = "Whether the user has been soft deleted or not", example = "false")
+	private boolean deleted;
 
 	public UserDTO() {
 	}
 
 	public UserDTO(String userName, UserGroupDTO userGroupName, String passwd, String desc) {
+		this(userName, userGroupName, passwd, desc, false);
+	}
+
+	public UserDTO(String userName, UserGroupDTO userGroupName, String passwd, String desc, boolean deleted) {
 		this.userName = userName;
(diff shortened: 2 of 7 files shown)


dbmalkovsky (commented):
Why are there no tests exercising the deletion code?

src/main/java/org/isf/usergroups/dto/UserGroupDTO.java
```diff
@@ -31,36 +31,38 @@
...
+	List<PermissionDTO> permissions;
 	@NotNull
 	@Schema(description = "Name of the group (must be unique)", example = "labo", maxLength = 50)
 	private String code;
-
 	@Schema(description = "The description of the group", example = "Staff members working in the laboratory", maxLength = 128)
 	private String desc;
-
-	@Schema(description = "List of group's permissions")
-	List<PermissionDTO> permissions;
```
mwithi: why to switch order of this field in the code?

src/main/java/org/isf/usergroups/dto/UserGroupDTO.java
```diff
@@ -72,4 +74,11 @@ public List<PermissionDTO> getPermissions() {
 	public void setPermissions(List<PermissionDTO> permissions) {
 		this.permissions = permissions;
 	}
+
+	public boolean isDeleted() {
+		return deleted;
+	}
+	public void setDeleted(boolean deleted) {
+		this.deleted = deleted;
+	}
```
mwithi:
```suggestion
	}
	
	public void setDeleted(boolean deleted) {
		this.deleted = deleted;
	}
```

src/main/java/org/isf/users/dto/UserDTO.java
```diff
@@ -78,4 +85,11 @@ public String getDesc() {
 	public void setDesc(String desc) {
 		this.desc = desc;
 	}
+
+	public boolean isDeleted() {
+		return deleted;
+	}
+	public void setDeleted(boolean deleted) {
+		this.deleted = deleted;
+	}
```
mwithi:
```suggestion
	}
	
	public void setDeleted(boolean deleted) {
		this.deleted = deleted;
	}
```

src/main/java/org/isf/users/rest/UserController.java
```diff
@@ -43,215 +45,221 @@
...
+	@ResponseStatus(HttpStatus.NO_CONTENT)
+	@DeleteMapping(value = "/users/{username}", produces = MediaType.APPLICATION_JSON_VALUE)
+	public void deleteUser(@PathVariable String username) throws OHServiceException {
+		User foundUser = userManager.getUserByName(username);
+		if (foundUser == null) {
+			throw new OHAPIException(new OHExceptionMessage("User not found."), HttpStatus.NOT_FOUND);
+		}
+		try {
+			try {
+				userManager.deleteUser(foundUser);
```
mwithi: This code was already in a try/catch block lines 181:188

mwithi: sorry, seen now that there is a further call to the manager

src/main/java/org/isf/users/rest/UserController.java
```diff
@@ -43,215 +45,221 @@
...
+			} catch (OHServiceException ex) {
+				foundUser.setDeleted(true);
+				userManager.deleteUser(foundUser);
+			}
```
mwithi: should API not doing it automatically but just report with an error and then the UI send a DTO already set for Deleted=True?

src/main/java/org/isf/usergroups/dto/UserGroupDTO.java
SteveGT96: It's the code style doing that ! I've just reformat the whole file using the IDE.

mwithi: I see... maybe because it is missing the `private` modifier so it fallsback to the default `protected` and therefore a different order? could you try please? 🤓

src/main/java/org/isf/users/rest/UserController.java
SteveGT96: Up to two request ? It could add an unncessary complexity level and it's error prone. By doing that, you should also send an explicit code to let the frontend know that it's a containt violation error. Does it really worth ?

mwithi: I think that anyway you have to inform the client about what is happening and why a user is hard-deleted and another might be only soft-delete. If I think about this, I easily end up with having the same logic as GUI

src/main/java/org/isf/usergroups/dto/UserGroupDTO.java
SteveGT96: Doen't change anything after reformating with the private modifier. I've copied it manually.

src/main/java/org/isf/users/rest/UserController.java
SteveGT96: Since on the ui we could not rely on the error messsage due to the fact that it is not yet standardized, we could notify the user before the deletion that the target object will be just marked as deleted if he's used somewhere. What do you think ?

mwithi: ok, but add a comment to improve that logic when the errors will be standardized

mwithi (changes requested):
Please fix tests

mwithi (approved):

src/main/java/org/isf/users/rest/UserController.java
```diff
@@ -43,215 +45,220 @@
...
+	public void deleteUser(@PathVariable String username) throws OHServiceException {
+		User foundUser = userManager.getUserByName(username);
+		if (foundUser == null) {
+			throw new OHAPIException(new OHExceptionMessage("User not found."), HttpStatus.NOT_FOUND);
+		}
+		try {
+			try {
+				userManager.deleteUser(foundUser);
+			} catch (OHServiceException ex) {
+				foundUser.setDeleted(true);
```
dbmalkovsky: This looks strange.  Why would I expect a failed `deleteUser()` method to suddenly succeed by setting the `isDeleted` field? 

I would expect lines 181-183 to be removed along with line 186.

So it would look something like the old code deleted above with a `foundUser.setDeleted(true);` call after the old line 176

dbmalkovsky (commented):
Added comments.

Also I don't see any new tests for soft deletion especially when the object is not found, and/or already deleted

src/main/java/org/isf/users/rest/UserController.java
mwithi: Hello @dbmalkovsky the aim, at this stage, is to try an hard-deletion and, if not possible, to proceed with a soft-deletion, in order to avoid two different API calls... this is notified to the user in the UI with a message before the call (e.g. "If the user cannot be deleted will be marked as "deleted". Ok to proceed?".
It is not perfect, I agree that we should have two different calls (one for hard-deletion and another one for update/soft-deletion, simply) but we can manage it in a later stage if it makes sense... wdyt?

dbmalkovsky: @mwithi OK.  I would vote to take the time to do it correctly now but I guess I can see the need to do it this way for now.   Just seems less professional to me.

mwithi: @SteveGT96 don't you think we should use the same pattern as in GUI?

delete -> only delete
update -> only update (soft-deletion is an update)

SteveGT96: > @SteveGT96 don't you think we should use the same pattern as in GUI?
> 
> delete -> only delete update -> only update (soft-deletion is an update)

Actually we are using almost the same pattern on the GUI. When we cannot delete the item due to DataIntegrityException, we make a second request to update it and mark it as deleted.

Please, could you describe very well what you want to be done ? It would be nice to describe also acceptance cases.

mwithi: Well, at the moment you don't actually do a second request but you just manage the update within the same call.
The standard procedure should be that, having the error message in return, the UI makes another call to the update method. If you want to make things consistent it shoud be like that, wdyt?

SteveGT96: I'd not be evident to rely on the error message to trigger the second call(since the request could faild for another reason and the error messages are not really standardized). That said, I suggest to split that in two actions on the frontend:

- "Delete":  for hard deletion.
- "Mark as deleted": that just update the item and mark it as deleted.

Does it look OK ?

mwithi: I think it's ok, I didn't mean to trigger the second call automatically, but managed by the user,  for which the update method should be enough

mwithi:
please resolve conflicts

mwithi:
I think checks are failing because API and CORE are not equally aligned with develop

mwithi (approved):
LGTM. Thanks a lot!

PR #506, opened 2024-11-19
OP-1365 Remove unnecessary static swagger-ui

mwithi:
See OP-1365.

Note: first commit pushed directly on develop by mistake.


dbmalkovsky (approved):
LGTM

PR #512, opened 2024-12-08
OP-1373 remove warns at api startup

mwithi:
See OP-1373.

Notes:
- we don't use [Mustache Templates](https://www.baeldung.com/mustache) and they are enabled by default in Spring Boot
- disabling `spring.jpa.open-in-view` is generally [recommended](https://www.baeldung.com/spring-open-session-in-view) for REST APIs to prevent performance issues and ensure explicit data fetching
- We don't need at the moment to define an authenticationProvider because we have only one UserDetailsService implementation (`org.isf.security.UserDetailsServiceImpl`) and only one PasswordEncoder bean (`BCryptPasswordEncoder`), so Spring Security uses the default DaoAuthenticationProvider. In the future, if we add more UserDetailsService implementations for multiple authentication data sources (e.g. Database, LDAP, Third-Party) we must define each provider and instruct the AuthenticationManager in which one to use for each case.

**NOTE**: `WARN - HHH90000025: MariaDBDialect ...` not eliminable because it is needed to start the application even without DB connection for the openapi workflow and in general

### DIFF
--- a/rsc/application.properties.dist
+++ b/rsc/application.properties.dist
@@ -38,4 +38,5 @@ jwt.token.secret=JWT_TOKEN_SECRET
 
 # Hibernate properties
 # needed to start application even without DB connection
+spring.jpa.database-platform=org.hibernate.dialect.MariaDBDialect
 spring.jpa.hibernate.ddl-auto=none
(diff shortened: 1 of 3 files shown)


dbmalkovsky (approved):
LGTM

dbmalkovsky (commented):
LGTM

PR #581, opened 2026-05-07
OP-1409 Remove duplicate AuditorAware bean from SecurityConfig

mwithi:
See OP-1409.

`spring.main.allow-bean-definition-overriding=true` is no longer needed

also removed obsolete SpringFox path matching workaround: `spring.mvc.pathmatch.matching-strategy=ANT_PATH_MATCHER`

Paired with https://github.com/informatici/openhospital-core/pull/1586.


dbmalkovsky (approved):
LGTM

PR #585, opened 2026-06-17
OP-896 Return mustChangePassword in the login response

giuseppesorge:
## OP-896 — Password lease (api)

Exposes the "must change password" state to API clients and enforces the policy server-side until the password is changed.

### Changes
- `LoginResponse` carries a new `mustChangePassword` flag, set at login when the user's must-change flag is on or the password lease has expired; `POST /auth/refresh-token` recomputes it from the database (it previously hardcoded `false`, silently disarming the web UI guard) and fails closed on a database error.
- **Server-side enforcement**: a new `MustChangePasswordFilter` (after `JWTFilter`) checks `passwdMustChange` and the lease expiration **from the database on every authenticated request** and rejects everything except `GET/PUT /users/me`, `POST /auth/login`, `POST /auth/logout`, `POST /auth/refresh-token` and `GET /healthcheck` with 403 until the password is changed. The JWT also carries a `mustChangePassword` claim for the client's convenience, but the server never trusts it (no stale-token window). The filter fails closed (403) on database errors and on tokens expiring mid-request.
- `PUT /users/me` rejects non-password profile updates with 403 while a password change is pending — the endpoint can only be used to change the password until the flag clears.
- Administrator-set passwords require a change at next login: password reset (`PUT /users/{username}`) flags the password as must-change; user creation inherits the flag from core (`UserBrowsingManager.newUser`, paired core PR); a user changing their own password (`PUT /users/me`) clears the flag.
- `POST /users` no longer echoes the (hashed) password in the 201 response body, aligning with the other endpoints.
- OpenAPI spec (`openapi/oh.yaml`) updated; verified the generated spec matches.
- Tests: filter scenarios (stale token, whitelist incl. login/healthcheck, cleared flag releases the same token, database error and mid-request expiry fail closed), ArgumentCaptor assertions on the flag propagation for admin reset and user creation, weak-password rejection on the admin paths, refresh fail-closed, no-password-echo on create.

### Known limitation (scope confirmation welcome)
Re-submitting the **identical** current password passes the strength checks, clears the flag and resets the lease clock without an actual credential rotation (pre-existing behaviour, not introduced here); with `STRONGPASSWORD=false` and no minimum length the strength check accepts trivial values. If old-vs-new comparison is in scope for OP-896 we can add it, otherwise it may deserve its own ticket.

Requires the paired core PR informatici/openhospital-core#1596 (same branch name, so CI cross-builds against it). Part of the OP-896 multi-repo change.

JIRA: https://openhospital.atlassian.net/browse/OP-896


giuseppesorge:
OP-896 PR set (all on branch `OP-896-password-lease`, base `develop`):
- core: informatici/openhospital-core#1596
- api: informatici/openhospital-api#585
- gui: informatici/openhospital-gui#2211
- ui: informatici/openhospital-ui#775
- doc: informatici/openhospital-doc#279

giuseppesorge:
Note for reviewers: the red CI is not caused by this change — the two failures (`AdmissionControllerTest.testDischargeAdmission_200` / `testNewAdmissions_201`) are the known OP-1424 mapper regression on develop, fixed by #605 (pending merge). This PR's own tests pass: TokenProvider (22), LoginController (7), MustChangePasswordFilter (10), UserController (21), JWTFilter (10) — all green locally on JDK 17.

giuseppesorge:
Update: #605 is merged — the branch now includes it (develop merged in `7b603c6`), so the CI should be back to green.

mwithi (approved):
LGTM. Thanks a lot!

PR #589, opened 2026-06-19
OP-891 Mask internal errors in API responses

giuseppesorge:
### Problem
`OHAPIError` serialized a **full Java stack trace** (`stackTrace`) and the raw exception messages (`debugMessage`) in **every** error response body, leaking exception class names, internal paths and SQL details to clients. (OP-890's `server.error.*` props don't help — the API serializes errors through this custom advice, bypassing Spring's error path.)

### Fix
- **`OHAPIError`**: drop `stackTrace` and `debugMessage`; expose only `status` + a clean user-facing `message` (plus `timestamp`/`description`). Add a `(HttpStatus, String)` constructor.
- **`OHResponseEntityExceptionHandler`**: add a catch-all `@ExceptionHandler(Exception.class)` → HTTP 500 `{"message":"An internal error occurred"}`, logging the full exception (with trace) to the application log only. Existing specific handlers (validation/404/OHServiceException) still win and keep their user-facing messages.
- New `OHResponseEntityExceptionHandlerTest` asserting error bodies never contain stack traces, exception class names or internal/SQL details (plants a fake SQL secret and asserts its absence).

### ⚠️ API response-contract change
The `stackTrace` and `debugMessage` JSON fields are **no longer returned** in error responses (intentional — this is the disclosure fix). No API-side or UI consumer reads them (only MSW mock fixtures referenced `debugMessage`).

### Verification
- Full API suite **229** green + new advice test (2). OpenAPI spec (`oh.yaml`) byte-equal (`OHAPIError` isn't in the spec).
- Live e2e: an OH-level error now returns `{status, message, timestamp, description}` only — no `stackTrace`/`debugMessage`.

GUI was reviewed and is **already compliant** (0 `printStackTrace`, errors funnel through localized `OHServiceExceptionUtil`) → no GUI change. The **openhospital-ui** part (React error boundary / sanitized client errors) is deferred to a separate ticket. Doc PR: _linked in a comment_.

### DIFF
--- a/src/main/java/org/isf/shared/exceptions/OHAPIError.java
+++ b/src/main/java/org/isf/shared/exceptions/OHAPIError.java
@@ -1,6 +1,6 @@
 /*
  * Open Hospital (www.open-hospital.org)
- * Copyright © 2006-2025 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
+ * Copyright © 2006-2026 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
  *
  * Open Hospital is a free and open source software for healthcare data management.
  *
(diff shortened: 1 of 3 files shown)


giuseppesorge:
Companion doc PR for OP-891: informatici/openhospital-doc#284

PR #592, opened 2026-06-23
OP-1037 Expose the administrative consensus flag in the REST API

giuseppesorge:
## OP-1037 — Administrative flag (API)

Exposes the administrative flag and its reason through the REST API, on both paths that carry the consensus.

The flag marks a patient whose administrative position has to be solved before they are served again; the reason is the optional free text explaining what. See the ticket for @mwithi's decisions on the semantics.

### Changes
- `PatientConsensusDTO`: `administrativeFlag` and `administrativeReason` (+ `@Schema`, accessors) — the dedicated consensus endpoint.
- `PatientDTO`: `consensusAdministrativeFlag` and `consensusAdministrativeReason` (+ `@Schema`, accessors), with the read and write ModelMapper mappings in `shared/mapper/mappings/PatientMapping`.
- `openapi/oh.yaml` updated for both DTOs. Both properties are optional, so this is an additive change and not breaking.

### Note for reviewers
`PUT /patients` stays a full replacement: a client that omits the flag clears it, exactly as it already does for the service consent flag. Worth a separate ticket if we want partial updates there.

### Depends on
- core informatici/openhospital-core#1620. CI here stays red until that is on `develop`.

### Testing
- Full api suite green (266 tests) against the companion core branch, including a round-trip test that checks both fields survive the mapping in both directions and that the administrative flag is not swapped with the service flag.

Part of OP-1037: core informatici/openhospital-core#1620, GUI informatici/openhospital-gui#2222, doc informatici/openhospital-doc#296, web UI informatici/openhospital-ui#817.


giuseppesorge:
Updated after @mwithi's design decisions on [OP-1037](https://openhospital.atlassian.net/browse/OP-1037): the administrative flag now carries an optional reason, exposed here on both `PatientConsensusDTO` and `PatientDTO` and mapped in both directions.

The OpenAPI change is additive (two optional properties), so it is not breaking. `develop` is merged in. Full api suite green (266 tests) against the companion core branch.

Note that `PUT /patients` stays a full replacement: a client that omits the flag clears it, exactly as it already does for the service consent flag. Happy to open a separate ticket if we want partial updates there.

[OP-1037]: https://openhospital.atlassian.net/browse/OP-1037?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #594, opened 2026-07-02
OP-1342 Add a JWT token revocation mechanism

giuseppesorge:
See [OP-1342](https://openhospital.atlassian.net/browse/OP-1342), part of the OP-1329 TokenProvider epic.

Light session-family revocation: at login a random UUID is stamped as the `jti` claim on both the access and the refresh token, and `/auth/refresh-token` propagates the same family id to each new pair (a legacy no-`jti` refresh token is adopted into a fresh revocable family). Logout reads the presented bearer token — valid or already expired (`ExpiredJwtException.getClaims()`), which is the common state at logout — and blacklists its family id in a new in-memory `TokenBlacklistService`. `TokenProvider.validateToken` returns the new `REVOKED` result for any token of a blacklisted family, `JWTFilter` answers 401, and the refresh endpoint rejects it through its existing `== VALID` guard with no controller change. One logout therefore kills the access token, the sibling refresh token and every pair refreshed from them, without touching other sessions of the same (possibly shared) account. Blacklist entries evict after the larger of the two configured validities; memory is bounded at one small entry per logout within that window, and only signature-verified tokens can create entries.

Trade-offs, disclosed:
- In-memory and per-instance: a JVM restart clears the blacklist, so previously revoked tokens work again until natural expiry (incident escape hatch: rotate `jwt.token.secret`). `TokenBlacklistService` is the single seam where a persistent or shared store could later replace the map.
- Tokens minted before this change carry no `jti` and stay unrevocable until they expire; each refresh of a legacy token mints a fresh revocable family, while the legacy refresh token itself remains unrevocable until expiry (at most the rememberMe validity).
- The React UI currently calls `POST /auth/logout` without the `Authorization` header, so browser logouts remain a logged no-op until the 1-line companion UI PR lands (linked below); clients that send their token get real revocation immediately.
- Pre-existing and unchanged: `/auth/refresh-token` does not distinguish access from refresh tokens (no `typ` claim). Post-logout the family check covers both; a `typ` claim could be a small follow-up under OP-1329.

Verified: full API suite green (254 tests) and live end-to-end against the demo database — login, protected call 200, logout 200, same access token 401 "JWT token has been revoked.", old refresh token rejected, immediate re-login works with a new family.

[OP-1342]: https://openhospital.atlassian.net/browse/OP-1342?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


giuseppesorge:
Companion UI PR: informatici/openhospital-ui#777

giuseppesorge:
Merged `develop` (1c6750e4). This was a real conflict: this PR and OP-896 (#585, now on develop) both added a third parameter to `TokenProvider.generateJwtToken` — a token family id here, `mustChangePassword` there. They are combined rather than chosen between: the implementation is now `generateJwtToken(Authentication, boolean rememberMe, boolean mustChangePassword, String tokenFamilyId)`, with the shorter overloads delegating to it and minting a fresh family id.

`LoginController` keeps both behaviours, on login and on refresh: the family id is reused (or a legacy token without one is adopted into a new, revocable family) and `mustChangePassword` is recomputed on every refresh. The stubs and verifications of both sides' tests were updated accordingly — `LoginControllerTest` and `TokenProviderTest` run green, 41 tests on JDK 17.

PR #595, opened 2026-07-02
OP-887 Add GET /patients/{code}/export for GDPR data portability

giuseppesorge:
See [OP-887](https://openhospital.atlassian.net/browse/OP-887). Depends on informatici/openhospital-core#1633 (same branch name, so this PR's CI cross-builds against that core branch).

`GET /patients/{code}/export` returns the full patient aggregate as JSON; with `Accept: text/csv` it returns a CSV rendering — one RFC-4180 section per entity group with `# <group>` marker lines, CRLF, and an attachment `Content-Disposition` (`patient_<code>_export.csv`). Media types are evaluated in header order, JSON winning when both are acceptable (q-values are not weighted; documented in the javadoc). The endpoint requires the new `patient.export` authority: the SecurityConfig rule is placed before the generic `GET /patients/**` matcher. `@Operation(description = "GDPR Art. 20 - Right to data portability")` per the ticket.

`openapi/oh.yaml` was regenerated and verified byte-identical (sorted-JSON compare) to the springdoc output of the running application.

This PR also fixes a pre-existing production bug surfaced while wiring the mapper: the shared ModelMapper mis-resolved `PatientVaccine.vaccine` onto `vaccineDate`, so `GET/POST/PUT /patientvaccines` failed with a MappingException for any record with a non-null vaccine date. Pinned with an explicit typeMap in `PatVacMapper`; an audit of other implicit GenericMapper matches may be worth a follow-up ticket.

Verified: controller and CSV tests green, full suite green (237 tests); live end-to-end on the demo database — 200 JSON with all entity groups populated (vaccines included, per the ticket's note), 200 CSV, 404 for an unknown patient, 403 for a user whose group lacks the permission.

[OP-887]: https://openhospital.atlassian.net/browse/OP-887?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/openapi/oh.yaml
+++ b/openapi/oh.yaml
@@ -767,6 +767,31 @@ paths:
                 type: boolean
       security:
       - bearerAuth: []
+  /patients/{code}/export:
+    get:
+      tags:
+      - Patients
+      description: GDPR Art. 20 - Right to data portability
+      operationId: exportPatientData
+      parameters:
+      - name: code
+        in: path
+        required: true
+        schema:
+          type: integer
+          format: int32
+      responses:
+        "200":
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/PatientExportDTO"
+            text/csv:
+              schema:
+                type: string
+      security:
+      - bearerAuth: []
   /patientconsensus/{patientId}:
     get:
       tags:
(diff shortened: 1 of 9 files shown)


giuseppesorge:
Hardened the CSV export against **formula injection** (commit `a0ba653`).

`PatientExportCsv.escape()` already applied RFC 4180 quoting, but a cell whose value starts with `=`, `+`, `-`, `@`, a tab or a carriage return is interpreted as a *formula* by spreadsheet applications (Excel, LibreOffice, Google Sheets) when the exported CSV is opened — a classic CSV/formula-injection vector, and here the exported fields come from user-controlled patient data.

The fix prefixes any such value with a single quote (`'`) before the RFC 4180 quoting. It lives inside `escape()`, so it covers **every cell** — both the header keys and the values, including the values produced by `render()` for nested JSON objects. Added a `neutralizesCsvFormulaInjection` unit test; `PatientExportCsvTest` is green (3/3).

Context: flagged by the automated security review of the pushed commits (MEDIUM). This only affects the CSV variant of the OP-887 export; the JSON body and the OP-888 full-record view (JSON only) are unaffected.

PR #596, opened 2026-07-02
OP-1244 Improve test coverage for org.isf.agetype.rest

giuseppesorge:
See [OP-1244](https://openhospital.atlassian.net/browse/OP-1244) (parent OP-1237: bring each listed package to at least 80% line coverage).

Improves line coverage of `org.isf.agetype.rest` from **70.6% to 100%** (branches 10/10, measured with jacoco 0.8.12) by extending the existing `AgeTypeControllerTest` with 7 tests covering the previously untested error paths of `PUT /agetypes` (null/empty/unknown code → 400, lookup failure and manager update failure → 500) and the miss branches of the age-code lookups (`GET /agetypes/code` returning no match, GET by index returning null → 404).

Note: the no-matching-age test pins the controller's current behavior of returning 200 with an empty JSON body rather than 404 — the test documents existing behavior rather than endorsing it. Only the test file is touched; 11/11 tests green.

[OP-1244]: https://openhospital.atlassian.net/browse/OP-1244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/test/java/org/isf/agetype/rest/AgeTypeControllerTest.java
+++ b/src/test/java/org/isf/agetype/rest/AgeTypeControllerTest.java
@@ -1,6 +1,6 @@
 /*
  * Open Hospital (www.open-hospital.org)
- * Copyright © 2006-2025 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
+ * Copyright © 2006-2026 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
  *
  * Open Hospital is a free and open source software for healthcare data management.
  *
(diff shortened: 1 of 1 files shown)


giuseppesorge:

Applied here in `ba5d9d8` — 7 method(s) renamed. Only the ones this branch introduces are touched; the `testXxx` methods already on `develop` are left as they are, so the diff stays about this PR. Suite re-run on JDK 17, still green.

giuseppesorge:
Following the review on #604 (https://github.com/informatici/openhospital-api/pull/604#discussion_r3677941770): the 7 tests this branch adds ended with `LOGGER.debug("result: {}", result)`, which asserts nothing — the test passed whatever the endpoint put in the body. They now assert the body itself (`54e1f44`). Two of them asserted only `containsString("is not valid.")`, which never checked which code the API rejected; they now match the full message.

Error paths send an explicit `Accept: application/json`, without which the error bodies negotiate to XML and no JSON field can be read. Suite green on JDK 17: 11 tests.

PR #597, opened 2026-07-02
OP-1245 Improve test coverage for org.isf.disease.rest

giuseppesorge:
See [OP-1245](https://openhospital.atlassian.net/browse/OP-1245) (parent OP-1237: bring each listed package to at least 80% line coverage).

Improves line coverage of `org.isf.disease.rest` from **68.8% to 100%** (branches 28/28, jacoco 0.8.12) by adding 17 tests to the existing `DiseaseControllerTest`: OHServiceException → 500 paths for all nine GET endpoints, 404 paths for get-by-code/update/delete, duplicate-code and duplicate-description → 400 plus not-created → 500 for POST, not-updated → 500 for PUT, and the delete-returns-false path. Every test asserts both the HTTP status and the OHAPIError message.

Also fixes a pre-existing copy-paste bug in `testGetDiseasesIpdInByCode_200`, which was hitting `/diseases/ipd/out/{typecode}` and mocking `getDiseaseIpdOut` instead of the ipd/in endpoint — this is why the ipd/in success path was uncovered; the fix is intentional, not scope creep.

Note: the delete-failure test pins the controller's current behavior of returning 200 with body `false` instead of an error status — documented as existing behavior. Only the test file is touched; 30/30 tests green.

[OP-1245]: https://openhospital.atlassian.net/browse/OP-1245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/test/java/org/isf/disease/rest/DiseaseControllerTest.java
+++ b/src/test/java/org/isf/disease/rest/DiseaseControllerTest.java
@@ -31,6 +31,7 @@ import static org.springframework.test.web.servlet.request.MockMvcRequestBuilder
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.put;
 import static org.springframework.test.web.servlet.result.MockMvcResultHandlers.log;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.content;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
 
 import java.util.List;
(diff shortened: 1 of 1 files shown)


giuseppesorge:

Applied here in `1f07a0c` — 17 method(s) renamed. Only the ones this branch introduces are touched; the `testXxx` methods already on `develop` are left as they are, so the diff stays about this PR. Suite re-run on JDK 17, still green.

giuseppesorge:
Following the review on #604 (https://github.com/informatici/openhospital-api/pull/604#discussion_r3677941770): the 17 tests this branch adds ended with `LOGGER.debug("result: {}", result)`, which asserts nothing — the test passed whatever the endpoint put in the body. They now assert the body itself (`67275b6`).

Error paths send an explicit `Accept: application/json`, without which the error bodies negotiate to XML and no JSON field can be read. Suite green on JDK 17: 30 tests.

PR #598, opened 2026-07-02
OP-1246 Improve test coverage for org.isf.distype.rest

giuseppesorge:
See [OP-1246](https://openhospital.atlassian.net/browse/OP-1246) (parent OP-1237: bring each listed package to at least 80% line coverage).

Improves line coverage of `org.isf.distype.rest` from **62.5% to 100%** (branches 6/6, jacoco 0.8.12) by adding 6 tests to the existing `DiseaseTypeControllerTest`: POST with duplicate code → 400, POST manager failure → 500, PUT with unknown code → 404, PUT manager failure → 500, DELETE with unknown code → 404, and DELETE where the manager throws → 200 with body `false`. Every test asserts both the HTTP status and the OHAPIError message.

Note: the last test pins the controller's current behavior of answering 200/`false` on a failed delete rather than an error status — the test documents existing behavior rather than endorsing it. Only the test file is touched; 10/10 tests green.

[OP-1246]: https://openhospital.atlassian.net/browse/OP-1246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/test/java/org/isf/distype/rest/DiseaseTypeControllerTest.java
+++ b/src/test/java/org/isf/distype/rest/DiseaseTypeControllerTest.java
@@ -32,6 +32,7 @@ import static org.springframework.test.web.servlet.request.MockMvcRequestBuilder
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.put;
 import static org.springframework.test.web.servlet.result.MockMvcResultHandlers.log;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.content;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
 
 import java.util.List;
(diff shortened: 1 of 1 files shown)


giuseppesorge:

Applied here in `0643d03` — 6 method(s) renamed. Only the ones this branch introduces are touched; the `testXxx` methods already on `develop` are left as they are, so the diff stays about this PR. Suite re-run on JDK 17, still green.

giuseppesorge:
Following the review on #604 (https://github.com/informatici/openhospital-api/pull/604#discussion_r3677941770): the 6 tests this branch adds ended with `LOGGER.debug("result: {}", result)`, which asserts nothing — the test passed whatever the endpoint put in the body. They now assert the body itself (`cb8eee3`).

Error paths send an explicit `Accept: application/json`, without which the error bodies negotiate to XML and no JSON field can be read. Suite green on JDK 17: 10 tests.

PR #599, opened 2026-07-02
OP-1274 Improve test coverage for org.isf.vaccine.rest

giuseppesorge:
See [OP-1274](https://openhospital.atlassian.net/browse/OP-1274) (parent OP-1237: bring each listed package to at least 80% line coverage).

Improves line coverage of `org.isf.vaccine.rest` from **69.0% to 100%** (branches 2/2, jacoco 0.8.12) by adding 5 tests to the existing `VaccineControllerTest` covering the previously untested error paths: POST with OHDataIntegrityViolationException → 400 "Vaccine type already present.", POST with OHServiceException → 400 "Vaccine not created.", PUT failure → 400 "Vaccine not updated.", DELETE of unknown code → 404 "Vaccine not found.", and DELETE where the manager throws → 400 "Vaccine not deleted.". Every test asserts both the HTTP status and the exact OHAPIError message.

Only the test file is touched; 11/11 tests green. No unreachable code remains in VaccineController.

[OP-1274]: https://openhospital.atlassian.net/browse/OP-1274?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/test/java/org/isf/vaccine/rest/VaccineControllerTest.java
+++ b/src/test/java/org/isf/vaccine/rest/VaccineControllerTest.java
@@ -22,6 +22,7 @@
 package org.isf.vaccine.rest;
 
 import static org.hamcrest.Matchers.containsString;
+import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.when;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.delete;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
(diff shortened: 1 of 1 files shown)


giuseppesorge:

Applied here in `5f31468` — 5 method(s) renamed. Only the ones this branch introduces are touched; the `testXxx` methods already on `develop` are left as they are, so the diff stays about this PR. Suite re-run on JDK 17, still green.

giuseppesorge:
Following the review on #604 (https://github.com/informatici/openhospital-api/pull/604#discussion_r3677941770): the 5 tests this branch adds ended with `LOGGER.debug("result: {}", result)`, which asserts nothing — the test passed whatever the endpoint put in the body. They now assert the body itself (`be77180`).

Error paths send an explicit `Accept: application/json`, without which the error bodies negotiate to XML and no JSON field can be read. Suite green on JDK 17: 11 tests.

PR #601, opened 2026-07-05
OP-1262 Improve test coverage for org.isf.opetype

giuseppesorge:
## OP-1262 — Improve test coverage for `org.isf.opetype`

Adds a `MockMvc` standalone test for `OperationTypeController`, following the pattern of the existing type-controller tests (`AdmissionTypeControllerTest`, `AgeTypeControllerTest`, `DiseaseTypeControllerTest`, …).

- **`OperationTypeControllerTest`** — covers all four endpoints:
  - `POST /operationtypes` → 201
  - `PUT /operationtypes/{code}` → 200
  - `GET /operationtypes` → 200
  - `DELETE /operationtypes/{code}` → 200
- **`OperationTypeDTOHelper`** — test-data helper built on the core `TestOperationType` fixture, mirroring the existing `*DTOHelper` classes.

The test wires the **real** `OperationTypeMapper` (not a mock) into the controller, so it exercises `org.isf.opetype.mapper` as well as `org.isf.opetype.rest`.

Jira: https://openhospital.atlassian.net/browse/OP-1262

### DIFF
--- a/src/test/java/org/isf/opetype/data/OperationTypeDTOHelper.java
+++ b/src/test/java/org/isf/opetype/data/OperationTypeDTOHelper.java
@@ -1,6 +1,6 @@
 /*
  * Open Hospital (www.open-hospital.org)
- * Copyright © 2006-2025 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
+ * Copyright © 2006-2026 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
  *
  * Open Hospital is a free and open source software for healthcare data management.
  *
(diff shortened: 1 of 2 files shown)


giuseppesorge:

Applied here in `bbcb3c7` — 4 method(s) renamed. Only the ones this branch introduces are touched; the `testXxx` methods already on `develop` are left as they are, so the diff stays about this PR. Suite re-run on JDK 17, still green.

giuseppesorge:
Following the review on #604 (https://github.com/informatici/openhospital-api/pull/604#discussion_r3677941770): all 4 tests ended with `LOGGER.debug("result: {}", result)`, which asserts nothing — the test passed whatever the endpoint put in the body. They now assert the body itself (`12bd97b`), so the `LOGGER` field and the `MvcResult` local are gone along with the redundant `is2xxSuccessful()` matchers.

This file is new in this PR, so the whole of it is converted; the repo-wide sweep over the tests already on develop is **OP-1452** (https://openhospital.atlassian.net/browse/OP-1452), filed off that thread. The 404 paths send an explicit `Accept: application/json`, without which the error bodies negotiate to XML and no JSON field can be read. Suite green on JDK 17: 4 tests.

PR #603, opened 2026-07-09
OP-1258 Add test coverage for the medical type REST layer

giuseppesorge:
Sub-task of OP-1237 (Testing Coverage Improvements - API): https://openhospital.atlassian.net/browse/OP-1258

Adds test coverage for the `org.isf.medtype` REST layer: `MedicalTypeControllerTest` (MockMvc standalone) covering all `MedicalTypeController` endpoints — list, create (201), update (200 and 404 when the code is missing), code check, and delete (200 and 404 when the code is not found) — plus a `MedicalTypeHelper` for the test data, following the existing `*TypeControllerTest` convention.

### DIFF
--- a/src/test/java/org/isf/medtype/data/MedicalTypeHelper.java
+++ b/src/test/java/org/isf/medtype/data/MedicalTypeHelper.java
@@ -1,6 +1,6 @@
 /*
  * Open Hospital (www.open-hospital.org)
- * Copyright © 2006-2025 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
+ * Copyright © 2006-2026 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
  *
  * Open Hospital is a free and open source software for healthcare data management.
  *
--- a/src/test/java/org/isf/medtype/data/MedicalTypeHelper.java
+++ b/src/test/java/org/isf/medtype/data/MedicalTypeHelper.java
@@ -1,0 +1,81 @@
+/*
+ * Open Hospital (www.open-hospital.org)
+ * Copyright © 2006-2025 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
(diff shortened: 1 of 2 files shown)


giuseppesorge:
The failing `build` check here is **not** caused by this PR — the added `MedicalTypeControllerTest` passes in CI (`Tests run: 7, Failures: 0`). The 2 failures are the pre-existing `AdmissionControllerTest.testNewAdmissions_201` / `testDischargeAdmission_200`, a `develop` regression from OP-1005: `Admission` now extends `SoftDeletableAuditable`, so ModelMapper maps the DTO's `deleted` from the new `deletedBy`/`deletedDate` fields and leaves it null → 400. Tracked as OP-1424 and fixed in #605; this PR will go green once #605 is merged.

cc @mwithi @dbmalkovsky

giuseppesorge:

Applied here in `70e09af` — 7 method(s) renamed. Only the ones this branch introduces are touched; the `testXxx` methods already on `develop` are left as they are, so the diff stays about this PR. Suite re-run on JDK 17, still green.

giuseppesorge:
Following the review on #604 (https://github.com/informatici/openhospital-api/pull/604#discussion_r3677941770): all 7 tests ended with `LOGGER.debug("result: {}", result)`, which asserts nothing — the test passed whatever the endpoint put in the body. They now assert the body itself (`a42705e`), so the `LOGGER` field and the `MvcResult` local are gone along with the redundant `is2xxSuccessful()` matchers.

This file is new in this PR, so the whole of it is converted; the repo-wide sweep over the tests already on develop is **OP-1452** (https://openhospital.atlassian.net/browse/OP-1452), filed off that thread. The 404 paths send an explicit `Accept: application/json`, without which the error bodies negotiate to XML and no JSON field can be read. Suite green on JDK 17: 7 tests.

PR #604, opened 2026-07-09
OP-1267 Add test coverage for the pregnant treatment type REST layer

giuseppesorge:
Sub-task of OP-1237 (Testing Coverage Improvements - API): https://openhospital.atlassian.net/browse/OP-1267

Adds test coverage for the `org.isf.pregtreattype` REST layer: `PregnantTreatmentTypeControllerTest` (MockMvc standalone) covering all endpoints — list, create (201), update (200 and 404 when the code is not present) and delete (200 and 404 when the code is not found) — and extends the existing `PregnantTreatmentTypeHelper` with the JSON serialization helpers used by the controller test.

### DIFF
--- a/src/test/java/org/isf/pregtreattype/rest/PregnantTreatmentTypeControllerTest.java
+++ b/src/test/java/org/isf/pregtreattype/rest/PregnantTreatmentTypeControllerTest.java
@@ -1,0 +1,220 @@
+/*
+ * Open Hospital (www.open-hospital.org)
+ * Copyright © 2006-2025 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
+ *
+ * Open Hospital is a free and open source software for healthcare data management.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * https://www.gnu.org/licenses/gpl-3.0-standalone.html
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+package org.isf.pregtreattype.rest;
+
+import static org.hamcrest.Matchers.containsString;
+import static org.mockito.Mockito.when;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.delete;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.put;
+import static org.springframework.test.web.servlet.result.MockMvcResultHandlers.log;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.content;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+import java.util.List;
+import java.util.Objects;
+
+import org.isf.pregtreattype.data.PregnantTreatmentTypeHelper;
+import org.isf.pregtreattype.dto.PregnantTreatmentTypeDTO;
+import org.isf.pregtreattype.manager.PregnantTreatmentTypeBrowserManager;
+import org.isf.pregtreattype.mapper.PregnantTreatmentTypeMapper;
+import org.isf.pregtreattype.model.PregnantTreatmentType;
+import org.isf.shared.exceptions.OHResponseEntityExceptionHandler;
+import org.isf.shared.mapper.converter.BlobToByteArrayConverter;
+import org.isf.shared.mapper.converter.ByteArrayToBlobConverter;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+import org.modelmapper.ModelMapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.http.MediaType;
+import org.springframework.test.util.ReflectionTestUtils;
+import org.springframework.test.web.servlet.MockMvc;
+import org.springframework.test.web.servlet.MvcResult;
+import org.springframework.test.web.servlet.setup.MockMvcBuilders;
+
+class PregnantTreatmentTypeControllerTest {
+
+	private static final Logger LOGGER = LoggerFactory.getLogger(PregnantTreatmentTypeControllerTest.class);
+
+	@Mock
+	protected PregnantTreatmentTypeBrowserManager pregnantTreatmentTypeManager;
+
+	protected PregnantTreatmentTypeMapper pregnantTreatmentTypeMapper = new PregnantTreatmentTypeMapper();
+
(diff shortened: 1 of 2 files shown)


giuseppesorge:
The failing `build` check here is **not** caused by this PR — the added `PregnantTreatmentTypeControllerTest` passes in CI (`Tests run: 6, Failures: 0`). The 2 failures are the pre-existing `AdmissionControllerTest.testNewAdmissions_201` / `testDischargeAdmission_200`, a `develop` regression from OP-1005: `Admission` now extends `SoftDeletableAuditable`, so ModelMapper maps the DTO's `deleted` from the new `deletedBy`/`deletedDate` fields and leaves it null → 400. Tracked as OP-1424 and fixed in #605; this PR will go green once #605 is merged.

cc @mwithi @dbmalkovsky

src/test/java/org/isf/pregtreattype/rest/PregnantTreatmentTypeControllerTest.java
```diff
@@ -0,0 +1,220 @@
...
+		ReflectionTestUtils.setField(pregnantTreatmentTypeMapper, "modelMapper", modelMapper);
+	}
+
+	@AfterEach
+	void closeService() throws Exception {
+		closeable.close();
+	}
+
+	@Test
+	void testGetAllPregnantTreatmentTypes_200() throws Exception {
```
dbmalkovsky:
```suggestion
	void getAllPregnantTreatmentTypes_200() throws Exception {
```
I know I can be a pain but typically the best practice for Junit 5 is to crop the 'test' prefix: In JUnit 5, the @Test annotation handles this context.

src/test/java/org/isf/pregtreattype/rest/PregnantTreatmentTypeControllerTest.java
```diff
@@ -0,0 +1,220 @@
...
+
+		MvcResult result = this.mockMvc
+			.perform(put(request, code)
+				.contentType(MediaType.APPLICATION_JSON)
+				.content(Objects.requireNonNull(PregnantTreatmentTypeHelper.asJsonString(body)))
+			)
+			.andDo(log())
+			.andExpect(status().is2xxSuccessful())
+			.andExpect(status().isOk())
+			.andReturn();
```
dbmalkovsky: instead of using debug to see the result (which doesn't verify anything in the code) how about adding something like:
<pre>
  .andExpect(MockMvcResultMatchers.jsonPath("$.code").value("ZZ"))
  .andExpect(MockMvcResultMatchers.jsonPath("$.description").value("TestDescription"))
  .andExpect(MockMvcResultMatchers.jsonPath("$.hashCode").value(0))
</pre>

src/test/java/org/isf/pregtreattype/rest/PregnantTreatmentTypeControllerTest.java
giuseppesorge: Not a pain at all — you're right, and it applies well beyond this file. Done in `c49728b`.

Since the same prefix is on every test I added in the sibling coverage PRs, I ran the same pass there too rather than leaving them inconsistent with this one:

| PR | commit | renamed |
|---|---|---|
| #596 OP-1244 | `ba5d9d8` | 7 |
| #597 OP-1245 | `1f07a0c` | 17 |
| #598 OP-1246 | `0643d03` | 6 |
| #599 OP-1274 | `5f31468` | 5 |
| #600 OP-1275 | `11314ac` | 5 |
| #601 OP-1262 | `bbcb3c7` | 4 |
| #603 OP-1258 | `70e09af` | 7 |

I deliberately renamed **only the methods those branches introduce**, and left the `testXxx` ones already on `develop` untouched — renaming those too would have buried the actual change under unrelated churn, which is the opposite of what you asked for on #2207. Say the word if you'd rather have the whole files aligned in one go, or a separate ticket for a repo-wide sweep.

Each suite was re-run on JDK 17 after the rename: 11, 30, 10, 11, 10, 4, 7 — all green.

src/test/java/org/isf/pregtreattype/rest/PregnantTreatmentTypeControllerTest.java
giuseppesorge: Agreed — the `LOGGER.debug` asserted nothing, it just made the test *look* like it checked something. Applied in `c49728b`, and to every test in the file rather than only this one, since they all had it.

Writing the assertions you suggested surfaced two things I would otherwise have gotten wrong:

**1. The error bodies are XML, not JSON.** `jsonPath("$.message")` fails on the 404s with `Expected to find an object with property ['message'] ... but found 'java.lang.String'`. The actual body is:

```
Content type = application/xml;charset=UTF-8
Body = <OHAPIError><status>NOT_FOUND</status><message>Pregnant Treatment Type not found.</message>...
```

The controller declares `produces = APPLICATION_JSON_VALUE`, but the `@RestControllerAdvice` returns its own `ResponseEntity<Object>`, which isn't covered by that. With no `Accept` header on the request, content negotiation picks XML. The tests now send `Accept: application/json` explicitly, which is what a real client does, and then `$.message` works.

**2. The "not created" branch answers 400, not 500** — `OHAPIException` defaults to `HttpStatus.BAD_REQUEST` and the controller doesn't override it there. I added a test pinning that, since it was uncovered.

So the file now asserts, per endpoint: the three fields you listed on POST/PUT, `hasSize(3)` plus the element fields on the list, `content().string("true")` on DELETE, and `$.message` on both 404s. 7/7 green on JDK 17.

One thing worth your call: `LOGGER.debug("result: {}", result)` is the established pattern in this repo — **131 occurrences across 20 test files on `develop`**, which is where I copied it from. I've only cleaned it up here. Happy to open a separate ticket to do the same across the rest if you want it gone repo-wide.

dbmalkovsky: Yeah I know the LOGGER pattern is used and I never liked it.   But I never took the time to research the "proper" way given I don't do much in the API project normally.

Yes I would add a new issue to make similar changes across the repo otherwise the code coverage numbers might look good but are they really accomplishing anything with validation?

giuseppesorge: Thanks for calling this out — you were right to, and the question about whether the coverage numbers accomplish anything was the useful part: my own tests were doing exactly what you described.

Filed as **OP-1452** (sub-task of OP-1237): https://openhospital.atlassian.net/browse/OP-1452

Scope measured on develop: 145 `LOGGER.debug` calls across 22 controller test files, 134 of them the `LOGGER.debug("result: {}", result)` idiom. The issue carries the per-response-type recipe, the two gotchas this PR turned up (error bodies negotiating to XML, `OHAPIException` defaulting to 400) and the capture-don't-guess trick — swap `log()` for `print()`, read the real bodies out of `target/test-reports/`, then write the assertions against them. Please adjust the scope or the estimate if you had something different in mind.

It seemed wrong to have the sibling PRs merge code the new issue would immediately have to clean up, so I ran the same pass over the tests my branches introduce; the pre-existing develop tests in those files are left for OP-1452.

| PR | commit | tests asserted |
|---|---|---|
| #596 OP-1244 | `54e1f44` | 7 |
| #597 OP-1245 | `67275b6` | 17 |
| #598 OP-1246 | `cb8eee3` | 6 |
| #599 OP-1274 | `be77180` | 5 |
| #600 OP-1275 | `1a6ec77` | 5 |
| #601 OP-1262 | `12bd97b` | 4 |
| #603 OP-1258 | `a42705e` | 7 |

All seven suites re-run green on JDK 17 (11 + 30 + 10 + 11 + 10 + 4 + 7).

One more thing your comment made me look at: where a body check did exist, it was usually `content().string(containsString(...))`, which passes wherever the text happens to appear in the payload. Two tests of mine in `AgeTypeControllerTest` asserted only `containsString("is not valid.")` — they never checked *which* code the API rejected. Those now match the full message via `jsonPath("$.message")`. Happy to revert any of it if you would rather keep the PRs to their original scope.

PR #605, opened 2026-07-09
OP-1424 Fix AdmissionDTO.deleted mapping regressed by the SoftDeletableAuditable change

giuseppesorge:
Fixes OP-1424: https://openhospital.atlassian.net/browse/OP-1424

Since OP-1005 made `Admission` extend `SoftDeletableAuditable` (adding the `String` fields `deletedBy`/`deletedDate`), ModelMapper — with ambiguity ignored — maps the DTO's `String deleted` from one of those new fields instead of the model's `char deleted`, leaving it null. `POST /admissions` and `PUT /admissions/{code}` then fail `@Valid` with **400**, so `AdmissionControllerTest.testNewAdmissions_201` and `testDischargeAdmission_200` fail on `develop` and redden the CI of **every open openhospital-api PR** (the api PR build runs against `develop`).

**Fix:** map `deleted` explicitly in both directions with a `char <-> String` converter (`AdmissionMapping`), wired via `@PostConstruct` in `AdmissionMapper` and in the controller test setup — the same pattern already used by `PatientMapping`. The full `openhospital-api` test suite is green (229 tests).

cc @mwithi @dbmalkovsky — this is a `develop` regression from OP-1005 (#1603 / core) that currently blocks the CI of all open api PRs.

### DIFF
--- a/src/main/java/org/isf/shared/mapper/mappings/AdmissionMapping.java
+++ b/src/main/java/org/isf/shared/mapper/mappings/AdmissionMapping.java
@@ -1,6 +1,6 @@
 /*
  * Open Hospital (www.open-hospital.org)
- * Copyright © 2006-2025 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
+ * Copyright © 2006-2026 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
  *
  * Open Hospital is a free and open source software for healthcare data management.
  *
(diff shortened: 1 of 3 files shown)


src/main/java/org/isf/shared/mapper/mappings/AdmissionMapping.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * Open Hospital (www.open-hospital.org)
+ * Copyright © 2006-2025 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
```
mwithi:
```suggestion
 * Copyright © 2006-2026 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
```

giuseppesorge: Done, thanks — fixed in the last commit.

mwithi (approved):
LGTM. Thanks a lot!

PR #608, opened 2026-07-16
OP-1431 Reject reusing the current password on a self password change

giuseppesorge:
## What

The self change-password endpoint (`PUT /users/me`) now rejects a new password equal to the current one, returning `400`. A new private `validatePasswordNotReused` mirrors the existing `validatePasswordStrength` and delegates to the core `UserBrowsingManager.isSameAsCurrentPassword`.

## Why

Follow-up to the password-policy work (OP-896 / OP-1430): a user renewing their own password must not be allowed to reuse the current one. The administrator reset path (`PUT /users/{username}`) is intentionally left unchanged, since an administrator does not know the target user's current password.

## Notes

- No change to the OpenAPI contract: no new endpoint or schema, only an added validation on an existing, already 400-capable endpoint.
- Controller test `shouldRejectReusingCurrentPassword`.

Depends on openhospital-core: OP-1431 (`isSameAsCurrentPassword`).

Jira: OP-1431

---
Related PRs: openhospital-core #1648 · openhospital-api #608 · openhospital-gui #2236

### DIFF
--- a/src/main/java/org/isf/users/rest/UserController.java
+++ b/src/main/java/org/isf/users/rest/UserController.java
@@ -1,6 +1,6 @@
 /*
  * Open Hospital (www.open-hospital.org)
- * Copyright © 2006-2025 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
+ * Copyright © 2006-2026 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
  *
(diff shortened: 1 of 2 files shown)


dbmalkovsky (approved):
LGTM; successfully tested with matching CORE PR

mwithi (approved):
Reviewed and approved by @dbmalkovsky

PR #586, opened 2026-06-18
OP-1216 Convert money and quantity DTO fields to DECIMAL/INT

giuseppesorge:
## OP-1216 — Align REST DTOs with the DECIMAL/INT type change

Aligns the API DTOs with the core type change (OP-1216).

### Changes
- **BigDecimal**: `BillDTO.amount/balance`, `BillItemsDTO.itemAmount`,
  `BillPaymentsDTO.amount`, `PriceDTO.price`, `MovementWardDTO.quantity`,
  `MedicalWardDTO.out_quantity`.
- **int**: `MedicalDTO` in/out/min quantity, `MedicalWardDTO.in_quantity`.
- Regenerated `openapi/oh.yaml` accordingly.

### Verification
- Full controller test suite green.
- OpenAPI gate replicated locally: ran the app, `mvn springdoc-openapi:generate`,
  and compared as sorted JSON (`yq`/`jq`) — the committed `oh.yaml` is byte-equal
  to the live-generated spec. The changes are additive/type-only (non-breaking).

Requires the core change. Companion PRs (same branch name) are linked in a comment below.

JIRA: https://openhospital.atlassian.net/browse/OP-1216

PR #612, opened 2026-07-28
OP-1445 Return stock movements instead of failing to map their dates

giuseppesorge:
Fixes [OP-1445](https://openhospital.atlassian.net/browse/OP-1445).

The core records a movement, and the lot it belongs to, with a date **and a time**, while the DTOs declared plain dates. ModelMapper's JSR-310 converter refuses to narrow `LocalDateTime` to `LocalDate`, so mapping the result list throws and every endpoint reading these DTOs answers 500:

```
Converter org.modelmapper.module.jsr310.TemporalToTemporalConverter
   failed to convert java.time.LocalDateTime to java.time.LocalDate.
Caused by: Unsupported mapping types[java.time.LocalDateTime->java.time.LocalDate]
```

That is `GET /stockmovements`, `/stockmovements/{ref}`, `/filter/v1`, `/filter/v2`, the four `medicalstockmovements` counterparts, and `GET /medicalstockward/movements/{ward_code}` and `/movements/to/{ward_code}` — the last two easy to miss, since their mandatory `from` and `to` keep them out of any plain sweep of the API.

`/stockmovements/lot/{med_code}` and its twin answer 200 today **only while they have nothing to return**: mapping a single lot through the production mapper fails the same way, which is verified in the tests. So the whole read surface is unusable and has been for some time.

## The two kinds of date

Per review — the endpoints have no consumers yet, so the request contract changes rather than being kept compatible by hand. `DateOrDateTimeDeserializer` is gone.

* **a movement is an event.** `MovementDTO.date` and `MovementWardDTO.date` are `LocalDateTime` and now accept only a date and time. Against real data the API returns movement dates with the time they were recorded at; a date-only field would drop it.
* **a lot's dates are dates.** `LotDTO.preparationDate` and `dueDate` go back to `LocalDate`. The time of day is never chosen for them — the GUI shows a date field and widens it internally — so exposing `2026-07-02T23:59:59` through the API added nothing a caller could use.

`LotMapping` converts explicitly in both directions: `toLocalDate()` on the way out, start of day for the preparation date and last second of the day for the due date on the way in, matching what `GoodDateChooser.getDateStartOfDay()`/`getDateEndOfDay()` hand the GUI. The model keeps `LocalDateTime`, so should the time of day ever become meaningful for lots, nothing has to move. The mapping is registered from `LotMapper`'s `@PostConstruct`, the way `AdmissionMapper` and `PatientMapper` already register theirs.

## The red `opeanpi` check

`build` is green, and inside `opeanpi` the step that checks the committed spec against the generated one passes — *No changes in OpenAPI spec*. What fails is the last step, the oasdiff breaking-change gate, and its three errors all come from one line, `MovementWardDTO.date` — the only one of these DTOs whose `@Schema` carries an explicit `format`:

```
POST /medicalstockward/movements                       the `date` request property `format` changed from `LocalDate` to `LocalDateTime`
GET /medicalstockward/movements/to/{target_ward_code}  the `items/date` response's property `format` changed ... for status `200`
GET /medicalstockward/movements/{ward_code}            the `items/date` response's property `format` changed ... for status `200`
```

`MovementDTO` is not flagged because it declares `type = "string"` and no `format`, so there is nothing in the spec for oasdiff to compare; `LotDTO` is no longer part of the change at all.

The gate cannot be satisfied while the field holds a `LocalDateTime`. Running oasdiff 1.26.1 — the version the action pins — over the two committed specs: `date-time` and `date` produce the same three errors, dropping the `format` attribute altogether produces two (the request one goes away, since a request that accepts more is not a break), and only leaving `LocalDate` produces none, which would mean labelling the field with a type it no longer has.

So the check is reporting exactly what it is there to report: a change of contract, accepted in review because these three operations answer 500 today and the API ships as EXPERIMENTAL. #586 (OP-1216, `double` → `BigDecimal`) was merged with this same check red, for the same class of change.

## Tests

* `StockMovementControllerTest` reproduces the failure without the fix, and asserts both halves of the contract: the movement keeps its time of day through the round trip, the lot's due date comes out as the day itself. It builds the mapper the way `ModelMapperConfig` does — with the JSR-310 module registered — because the offending conversion is only attempted once that module is present: a test written like the other controller tests here, with a bare `new ModelMapper()`, would pass either way.
* `MovementDateContractTest` pins the two shapes at the JSON boundary, and that a malformed value is still rejected on both.

Full suite green on JDK 17: 262 tests, 0 failures.

## Follow-up, not in this change

`GET /medicalstockward/movements/{ward_code}` still takes plain dates as `from` and `to` and converts them with `atStartOfDay()`, so its results are now finer-grained than the filter that selects them, and `to` excludes everything recorded on the last day itself. Both belong in a separate issue.

[OP-1445]: https://openhospital.atlassian.net/browse/OP-1445?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mwithi:
Since these endpoints do not have consumers yet (API=EXPERIMENTAL), I think we can accept the breaking request change and avoid adding `DateOrDateTimeDeserializer`.
`MovementDTO.date` and `MovementWardDTO.date` should become `LocalDateTime`, because a movement is an event and its time is meaningful. However, I would keep `LotDTO.preparationDate` and `LotDTO.dueDate` as `LocalDate`: the current GUI presents them as dates only and converts them to start-of-day and end-of-day internally, so exposing those artificial times through the API does not currently add useful information.
Could `LotMapper` explicitly convert the core values with `toLocalDate()` and map incoming preparation and due dates to start/end of day? This keeps the current core model unchanged, should time-of-day become relevant for lots in the future, while giving the API a simpler contract that reflects their present use.

giuseppesorge:
@mwithi done in `c2534e5` — taken as you described.

`DateOrDateTimeDeserializer` is gone. `MovementDTO.date` and `MovementWardDTO.date` stay `LocalDateTime` and now accept only a date and time. `LotDTO.preparationDate` and `dueDate` are back to `LocalDate`.

`LotMapping` does the conversion explicitly in both directions — `toLocalDate()` outbound, and inbound the preparation date to the start of its day and the due date to the last second of its day (`23:59:59`), which is what `GoodDateChooser.getDateStartOfDay()` and `getDateEndOfDay()` hand the GUI today, so a lot is not treated as expired throughout the day it expires on. The model is untouched, so nothing has to move if the time of day ever becomes meaningful for lots.

I put the converters in `org.isf.shared.mapper.mappings.LotMapping` and register them from `LotMapper`'s `@PostConstruct`, rather than inline in the mapper: that is the shape `AdmissionMapper`/`AdmissionMapping` and `PatientMapper`/`PatientMapping` already use here, and it is what lets a controller test register the same conversions on a hand-built `ModelMapper` — `StockMovementControllerTest` needs exactly that, since nothing calls `@PostConstruct` there.

`MovementDateCompatibilityTest` pinned the compatibility that is now deliberately dropped, so it is replaced by `MovementDateContractTest`, which pins the contract that holds: a movement carries its time, a lot carries plain dates, and a malformed value is rejected on both. `StockMovementControllerTest` now asserts both halves — the movement keeps `10:15:30` through the round trip, the lot's due date comes back as the day rather than `23:59:59`.

Full suite green on JDK 17: **262 tests, 0 failures**. `openapi/oh.yaml` is back to `develop`'s values for `LotDTO`; `MovementWardDTO.date` still changes `format`, so `opeanpi` stays red for the one reason described in the PR body — which I have updated, since the "nothing that works today stops working" framing no longer applies.

mwithi (approved):
LGTM. Thanks a lot!

